PR #36, opened 2015-01-06
add openaddresses type

missinglink:
add openaddresses type to schema


missinglink:
This is available from `pelias-schema@0.0.12`

PR #39, opened 2015-01-14
add osm address type

missinglink:
add `osmaddress` type to schema.

the idea here is that we start partitioning the address records from OSM in to their own elasticsearch `"type"`

related: https://github.com/pelias/openstreetmap/issues/29

cc/ @hkrishna @sevko @heffergm

### DIFF
--- a/schema.js
+++ b/schema.js
@@ -3,14 +3,15 @@ var schema = {
   'settings':           require('./settings')(),
   'mappings': {
 
-    // osm
+    // openstreetmap points-of-interest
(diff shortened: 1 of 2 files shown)


missinglink:
@heffergm FYI, this is another new ES 'type' called 'osmaddress'.

to add this to an existing index it's the same process as we did for the 'openaddresses' type, namely https://github.com/pelias/schema#reset-a-single-type.

otherwise it'll just be there when new indeces are built from scratch as-per-normal

PR #40, opened 2015-02-05
Adding multiplier type to POIs

hkrishna:
for storing population information when available (better scoring function)


missinglink:
LGTM, we can release this today @hkrishna if you're happy with it?

PR #41, opened 2015-02-17
Pelias ngram [DO NOT MERGE] [EXPERIMENT]

hkrishna:
This is purely experimental. I changed the schema to play with ngram tokenizer/token filters. Using ngram token filters posed a few problems including

```
[500] IllegalArgumentException[TokenStream expanded to 3114 finite strings. Only <= 256 finite strings are supported]
```
This happens because the ngram token filter expands a given token ```station``` to ```[s] [t] [a] [t] [i] [o] [n] [st] [ta] [at] [ti] [io] [on]``` which is okay but if you have a bigger token then it quickly gets out of hand.

So, I went back to using ngram tokenizer instead of token filters. 

@missinglink lets discuss ngrams here. I got it started for now. because I was hacking with it over the weekend - decided to push some of it here so, we can play with it - Made a new index called pelias_ngram and updated geonames (https://github.com/pelias/geonames/tree/using-pelias-ngram) and API (https://github.com/pelias/api/tree/using-pelias-ngram)

### DIFF
--- a/mappings/partial/suggest.json
+++ b/mappings/partial/suggest.json
@@ -2,8 +2,8 @@
   "type": "completion",
   "preserve_position_increments": false,
   "preserve_separators": false,
-  "index_analyzer": "plugin",
-  "search_analyzer": "plugin",
+  "index_analyzer": "suggestions_ngram",
+  "search_analyzer": "suggestions_ngram",
   "payloads": false,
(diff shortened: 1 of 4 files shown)


dianashk:
Waiting to see what ES 2.0.0 brings.

PR #42, opened 2015-02-18
change the way we analyze admin fields so we can search on them

hkrishna:
```keyword``` analyzer is useful for pin codes but for admin0 and admin1, we'd like to search for ```new york``` or  ```New York``` and always be able to match to ```New York``` using ```match``` query/filter. 

[this is a work in progress] - Problem with this PR is that ```'Great Britain'``` might get broken into ```['great', 'britain']``` 

Maybe all we need is a lowercase filter?? [Lets Discuss]

As per https://github.com/pelias/api/pull/68

Fixes pelias/pelias#43

### DIFF
--- a/mappings/partial/admin.json
+++ b/mappings/partial/admin.json
@@ -1,6 +1,6 @@
 {
   "type": "string",
-  "index_analyzer": "keyword",
-  "search_analyzer": "keyword",
+  "index_analyzer": "pelias",
+  "search_analyzer": "pelias",
   "store": "yes"
 }
--- a/test/partial-admin.js
+++ b/test/partial-admin.js
@@ -34,11 +34,12 @@ module.exports.tests.index = function(test, common) {
   });
 };
 
-// keyword analysis ensures that we get ['Great Britain'] instead of ['Great','Britain']
+// pelias analysis does not ensure that we get ['Great Britain'] instead of ['Great','Britain']
+// TODO this needs to be addressed 
 module.exports.tests.analysis = function(test, common) {
   test('index analysis', function(t) {
-    t.equal(schema.index_analyzer, 'keyword', 'should be keyword');
-    t.equal(schema.search_analyzer, 'keyword', 'should be keyword');
+    t.equal(schema.index_analyzer, 'pelias', 'should be pelias');
+    t.equal(schema.search_analyzer, 'pelias', 'should be pelias');
     t.end();
   });
 };


missinglink:
heya, is this a `DO NOT MERGE`?

I'm not sure if I understand what `'pin codes'` are?

Can you please bring this up on our next video call?

hkrishna:
This is definitely something that needs to get merged into master. The reason the PR is in a form of a ```discussion``` is because I wanted to understand the reasoning behind analyzing ```admin``` fields with ```keyword```. 

If you look at [keyword analyzer](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/analysis-keyword-analyzer.html) - it says that the field is not analyzed. In other words, if you have an admin value ```Flatiron```, the only way to search for documents with that admin value is to be case sensitive in our queries. So, if a user types in ```flatiron``` it wont match. 

By using ```pelias``` analyzer (or even a ```standard``` analyzer) we could make sure we add that flexibility which is much needed. Thoughts?

dianashk:
@hkrishna could you please add some acceptance tests that highlight the issues here. Would be good to run them on current build and then again once this is merged to master and quantify progress.

hkrishna:
Connects to pelias/api#68

hkrishna:
@dianashk sure, will do

missinglink:
Looking back over the history I saw the comment I put on the unit test when it was originally set to keyword.

Am I right it saying this could potentially have negative as well as positive impact?

Lowercase filtering sounds like it does what you want, the idea was to keep the 2 words as a single token, presumably because of undesirable results containing 'new' or 'great' when searching for 'New York' or 'Great Britain?

Having said that, we never searched on these fields in the past so if you are confident then feel free to ship it.

missinglink:
Just remembered, 'keyword' had to be used instead of 'standard' to get aggregations to work as per this example: https://gist.github.com/missinglink/87231ad6233fda54a8b3

We don't currently use this type of query anywhere, just a FYI on the history, if you try running that with 'standard' you can assertain my motivation for doing it like that.

hkrishna:
@missinglink yeah, that was the reason I first added the label 'question' to this PR. I agree that maybe ```lowercase``` is all we need here without a whitespace tokenizer. However, this is an improvement with what we have in master - there is no doubt we need to revisit this in the next iteration and tweak our analyzer to achieve the desired behavior. I've opened an issue https://github.com/pelias/schema/issues/46 for it.

Issue #48, opened 2015-03-27
rethink schema types


sevko:
[We currently define](https://github.com/pelias/schema/blob/master/schema.js#L7,L23) elasticsearch types on a per-dataset basis, so things like GeoNames, OpenAddresses, and OSM all have an individual type (OSM has multiple, in fact: `osmaddress`, `osmnode`, and `osmway`). The majority of these are all aliases for the [`poi` mapping](https://github.com/pelias/schema/blob/master/mappings/poi.js), which just defines a named point, which begs the question why we're bothering with a type per dataset to begin with. I believe it'd be better to have just, say, three types - `point` (addresses/POIs), `street`, and `boundary` (administrative polygons) - for the following reasons:

1. ease of use: this is really the biggest selling point. If someone wants to index a new dataset in Pelias, they either need to add a new type to the schema or use one of the existing ones. The problem here is that the user needs to *learn* about the different types, and we need to document that we have a type per dataset (and I don't think there's a clear reason why that is... we just *do*). People would then have to locally modify `pelias-schema` and recreate the index. Moreover, people won't *get* the different types right off the bat: what are `openaddresses`, `geonames`, `admin1`, etc? Then they have to go understand the POI mapping, and so on. It'd be much more intuitive for people to deal with types like `point` and `street`, because it should usually be painfully clear what kind of data you have, and those don't require much explanation.

2. clarity: this was touched on in the previous point, but the *type* of data seems like a much stronger cause for separation into different elasticsearch types than its *origin*. It makes sense to separate administrative polygons from addresses/POIs from street linestrings. I don't think it makes *as much* sense to segregate them by data-source, which seems rather secondary.

Here are the downsides:

1. we lose the ability to remove a specific dataset by just dropping a type. Kind of a non-issue, since we can just write an ES query to do the same thing and bundle that into a script.

2. we'll have to:
  1. change record GUIDs: we currently generate pseudo-GUIDs in a variety of ways that differ by dataset (some just use an incremented integer). We'd need to write a separate Transform stream that creates uniform ids for all datasets that are guaranteed to be unique in their elasticsearch type. I'm envisioning this as a separate step of the imports:
  
  ```javascript
  .pipe(peliasSuggester.pipeline)
  .pipe(peliasDocumentGuid()) // <--- record ID set here
  .pipe(peliasDbclient())
  ```

[...]

PR #55, opened 2015-05-12
simplify schema to use _default_ mapping

missinglink:
this PR aims to completely simplify the schema, using the `_default_` mapping on the `pelias` elasticsearch index means that all new `_type`'s created will inherit their mappings from `_default_`.

in effect this makes creating a new `_type` for a new dataset a no-brainer, completely removing the need for messing around in this repo to add the new mapping and then running a script.

I tested it out here: https://github.com/pelias/playground/blob/master/default_mapping.js#L80 and it behaved as expected, plus I did an OSM import and it worked fine, the `osm*` layers inherited their mappings from `_default_`


hkrishna:
looks good :+1:

dianashk:
:+1:

missinglink:
resolves https://github.com/pelias/schema/issues/48

missinglink:
```bash
$ npm publish
npm http PUT https://registry.npmjs.org/pelias-schema
npm http 200 https://registry.npmjs.org/pelias-schema
+ pelias-schema@0.4.0
```

PR #60, opened 2015-06-12
Set maximum input length for the suggester

### LINKED ISSUE #21: max_input_length
Explicitly define `max_input_length` for the suggester (the default of 50 is probably fine)

http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-suggesters-completion.html#completion-suggester-mapping

bradh:
Resolves #21.

Includes simple test.


missinglink:
Thanks @bradh!

PR #76, opened 2015-10-06
add source/layer/kind

missinglink:
this PR adds 3x new fields to the schema named:
- `source`
- `layer`
- `kind`

they all share the same analysis called `literal` which simply uses the `keyword` analyser.

when matching against the fields above you MUST specify the input text exactly as was entered in the field, so: 'Openstreetmap' !== 'openstreetmap'

this is the same type of analysis we use for `categories` so I simply renamed that analyser.

also added an end-to-end test


dianashk:
I'm not clear what `kind` is in reference to.

riordan:
`kind` is a part of the Mapzen terminology introduced with the [Vector Tiles](https://github.com/mapzen/vector-datasource#layers) that refers to the original OSM classification. @nvkelso's thoughts were that this would be where we would stuff the original OSM categories / where unnormalized categories would be when folks bring their own data.

missinglink:
`kind` is not a Pelias thing, its basically the 'main' category from the dataset it was derived.

in the near future we probably won't need to expose filter functionality, in the far future it would be handy to do so, this will be helpful to advanced users when the pelias category mapping is insufficient

An example would be to filter by osm 'amenity', or if you are bringing your own data it would allow you to use labels from your own domain such as '5 stars' or whatever the business uses.

dianashk:
Got it. Thanks for clarifying.

dianashk:
Also :+1:

missinglink:
in light of todays discussion, we should remove `kind` and add `source_id`.

`source_id` will contain all information required to locate the record in the source dataset, but does not include the name of the source itself, eg. `source: "osm", source_id: "way:201"`

we may choose to re-introduce `kind` at a later date, pending further discussion about data taxonomy

missinglink:
@pelias/contributors please review changes in https://github.com/pelias/schema/commit/dc27bc8482809c0d6c19540e459b2298ad693ac2 (comment above) and feel free to merge

dianashk:
There is a failing test in Travis. :(

missinglink:
updated end-to-end test to reflect changes made in master

PR #83, opened 2015-12-14
remove stop words filter from ngrams analyzers to remedy autocomplete street suffix jitter

missinglink:
this PR disables the `stop_words` filter for two ngram analyzers ( `peliasOneEdgeGram` & `peliasTwoEdgeGram`).

**Note**: I am only going to comment on `peliasTwoEdgeGram` as `peliasOneEdgeGram` is currently only used for countries and not intended for use with address data.

prior to this PR, [these terms](https://github.com/pelias/schema/blob/65cad0053d5be824c2bfa8b5ea69d1a20dd99b06/street_suffix.js#L2-L25) were being removed from those indexes to improve query performance.

The motivation there was that common street suffixes such as `street` and `road` would create very large indexes for the grams `st` and `ro` respectively; which would result in slow performance on those indices at query-time.

**Note:** The token filters were applied to full words *before* they were cut in to ngrams; this means that if the token `street` appeared in the string (surrounded by whitespace) is was *completely removed* and therefore created no ngrams.

```
peliasTwoEdgeGram

stop_words filter enabled:
"XX square"  => [ 'xx' ]
"XX squa"    => [ 'xx', 'sq', 'squ', 'squa' ]
"XX sq"      => [ 'xx', 'sq' ]

stop_words filter disabled:
"XX square"  => [ 'xx', 'sq', 'squ', 'squa', 'squar', 'square' ]
"XX squa"    => [ 'xx', 'sq', 'squ', 'squa' ]
"XX sq"      => [ 'xx', 'sq' ]
```

the side effect was poor user-experience when autocompleting street suffixes as demostrated below:

ngram suffix jitter img

remark: in the image above you can see that the correct place is found using the terms `union sq` and `union square`; however all intermediate representations such as `union squa` fail.

remark: comparing the image to the analysis above you can see that the `ngram` matching **will not** actually match on the fully specified term `"union square"` at all**!!** This behaviour is due to the `peliasPhrase` analyzer, which matches the completeness of **word** tokens.

---

so the tl;dr is that this PR will remove the `stop_words` filter from ngrams analyzers, this will solve the UX issues outlined above at the expense of an increased index size and possibly decreased performance.

please review and comment, I will stage this once everyone has had some time to review it. I can try to do some performance tests on a full cluster to determine the extent of the performance impact, but making an accurate measurement there will be difficult without spinning up 2x full dev clusters with the same hardware.

if there is a performance degradation we may be able to smooth it over with longer TTLs in the edge cache.

**Note:** an additional task needs to be performed in order to create synonyms for street suffixes who's abbreviations are not a prefix of their value. eg:

```
'st' is a prefix of 'street' (first 2 letters)
'rd' is NOT a prefix of 'road' (it will require a synonym 'rd' to be added to its grams)
```


orangejulius:
I'd definitely like to see this get built on dev very soon. I'm almost certain it will help a lot, but there's only one way to know...

missinglink:
resolves https://github.com/pelias/api/issues/357

PR #84, opened 2015-12-14
add a list of 'safe' expansions

missinglink:
following on from the closing note in https://github.com/pelias/schema/pull/83, please see that PR for more info before commenting here.

**Note:** this branch now includes https://github.com/pelias/schema/pull/83 (merged in)
**Note:** this branch now includes https://github.com/pelias/schema/pull/85 (merged in)
**Note:** this branch now includes https://github.com/pelias/schema/pull/86 (merged in)
**Note:** this branch DOES NOT include https://github.com/pelias/schema/pull/88 (trying not to make it too complex)

this is a first pass at a list of street abbreviations which are relatively 'safe' to expand.

from the code comments:

```
/**
  a list of 'safe' street suffix expansions.

 this list should NOT include any values where the abbreviation is a prefix of
 the expanded form.

 EG. 'st' is a prefix of 'street' so it is not included here.
 EG. 'rd' is NOT a prefix of 'road' so it IS included here.

 the term 'safe' refers to whether the token may be expanded without causing
 other issues; in general ask yourself "if i expand `byu => bayou` will this cause
 issues with other tokens which *begin with byu?"

 EG. 'pr' is disabled as it would cause jitter when autocompleting any place
 name beginning with 'pr' such as 'princeton', on the second keypress the
 results would likely all contain names which are/begin with "pier".

 EG. 'ct' is disabled as expanding it to 'court' would possibly conflict with
 the state abbreviation for 'Connecticut'.

 please use judgement when adding new expansions as it may cause the 'jitter'
 behaviour as outlined in https://github.com/pelias/schema/pull/83
**/
```

**Note:** this PR is not complete as it needs tests and to be enabled for the appropriate indices.

Please feel free to comment on the methodology and the tokens I have chosen.


riordan:
Great idea. Might be worth having a "Safe Expansions" tool that looks for conflicts.

integration/analyzer_peliasAdmin.js
```diff
@@ -103,4 +103,4 @@ function simpleTokens( tokens ){
   return tokens.map( function( t ){
     return t.token;
   });
-}
\ No newline at end of file
+}
```
orangejulius: What? You've finally bowed down to the true UNIX ways of newline chars at the end of the last line of a file? there is hope

settings.js
```diff
@@ -109,9 +108,8 @@ function generate(){
       },
       "filter" : {
         "ampersand" :{
-          "type" : "pattern_replace",
```
orangejulius: oh jeez how did we not see the issue with this

integration/analyzer_peliasAdmin.js
missinglink: atom

settings.js
missinglink: yea classic case of unit testing tunnel vision. it did what I expected, but also did a bunch more!

PR #85, opened 2015-12-17
fix ampersand

missinglink:
fixes an error where the `and -> &` substitution was being done *inside* tokens.

```
disneyland -> disneyl&
new zealand -> new zeal&
```

fixed in this PR


orangejulius:
oh no. nice catch

PR #86, opened 2015-12-17
increase max gram token size

missinglink:
increase max ngram size.

tokens such as `grolmanstrasse` and `shuffleboard` were being truncated at 10 chars

### DIFF
--- a/integration/analyzer_peliasOneEdgeGram.js
+++ b/integration/analyzer_peliasOneEdgeGram.js
@@ -26,7 +26,7 @@ module.exports.tests.analyze = function(test, common){
     assertAnalysis( 'ampersand', 'a and b', ['a','&','b'] );
     assertAnalysis( 'ampersand', 'a & b', ['a','&','b'] );
     assertAnalysis( 'ampersand', 'a and & and b', ['a','&','b'] );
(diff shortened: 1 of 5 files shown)


orangejulius:
looks like a good idea. presumably the index will get a bit bigger, but not much, since few words are that big, right? any ideas on how to measure the impact?

riordan:
Maybe simulate an import of just quattroshapes and seeing how many places have token_len>10. The admin areas tend to have the most cascading impact on the index, so seeing how extensive those areas are could give us an order of magnitude estimation of just how much impact there could be.


On 12/17/15 09:33, Julian Simioni wrote:
> looks like a good idea. presumably the index will get a bit bigger, but
> not much, since few words are that big, right? any ideas on how to
> measure the impact?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pelias/schema/pull/86#issuecomment-165466771>.
>

-- 
David Riordan | Product Manager - Search | Mapzen
dave.riordan@mapzen.com | gpg 235D 9DC9 5EF6 277C
@riordan | +1 (203)-521-1222
https://mapzen.com | @mapzen

PR #88, opened 2015-12-21
Remove ordinals

peter-johnson:
enable `remove_ordinals` filter for many analyzers

```
'20th 21st 22nd 23rd' -> ['20','21','22','23']
```

was previously only enabled for `peliasStreet` and is now enabled for `peliasOneEdgeGram`, `peliasTwoEdgeGram` and `peliasPhrase` also.


orangejulius:
This is old, let's remake the PR if we want to look at this again.

Issue #89, opened 2015-12-21
review kstem for autocomplete


missinglink:
review `kstem` token filter for autocomplete analyses.

eg. 'walking' -> 'walk'
...but 'walki' => '??'

in cases where the stem is not strictly a prefix of the expansion, what happens then?

eg. 'peoples' -> 'person'
... but 'peop' -> '??'

riordan:
Great catch

riordan:
@orangejulius sez:
> Maybe this makes sense to run only when we've parse out that it's a category.

PR #93, opened 2016-01-07
Remove kstem analysis filter

orangejulius:
This PR completely removes the kstem analysis filter, just as an experiment to see what sort of effect it has on search quality.

_Update:_ after running this on dev, I now propose that we actually remove kstem. Full rationale below.

I was a bit amazed when I ran our acceptance tests and found that removing kstem slightly improves the results

screenshot from 2016-01-11 14-10-36

I've gone through all the tests to see where the differences are, and here are my conclusions. I'm _really_ looking for people to help by finding other cases that don't work as well on dev right now, as if there are any serious issues, we should hold off on merging this. That said I haven't really found any.

### [daly city](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=daly%20city)

This is the original query that got us looking at kstem. Daly is stemmed to dale which causes dale city to match above daly city

### [brooklyn](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=brookly)

While the full query for brooklyn works fine, there is jumpiness when using autocomplete. `brook` is replaced with `brk` by our synonym matching. However, kstem also replaces `brookly` with `brook`, so the jumpiness appears twice in the query.

### [statue of liberty](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=statue%20of%20liberty)

I haven't fully dug into this one, but there are only minor changes in the ordering of results, and I don't see any weird stemming behavior, so it might be a data issue.

### [101 saint mark place](http://pelias.github.io/compare#/v1/autocomplete%3Ftext=101%20saint%20mark%20pl)

This test either has a typo or was written assuming kstemming means plurals never matter (the street is called St. Mark's Place). It's a good example of some fragility that does come from not handling plurals. I think we can either build a very conservative stemmer or do some magic at the API level to fix this. If you do a query for [101 saint marks pl](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=101%20saint%20marks%20pl), everything works.

### [whole foods ny](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=whole%20foods%20n)
This test has [no location bias](https://github.com/pelias/acceptance-tests/blob/master/test_cases/address_parsing.json#L295) and is searching for a chain with [431](https://en.wikipedia.org/wiki/Whole_Foods_Market) locations across the US. The " ny" at the end isn't currently enough signal for Pelias to prefer New York locations, however since the priority threshold is 5 and there are lots of Whole Foods locations in New York state, it usually passes. Note that with kstemming turned off, it's now possible to search for places that [actually have "whole food" in the name](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=whole%20food), with kstemming, they are overwhelmed by Whole Foods.
  
  

### Stemming isn't really for geocoding

Stemming was designed so that information retrieval systems can find documents that pertain to a similar topic to some keywords. With geocoding, users are entering something's name. While we want to allow some leeway in the name, the strategies for doing so seem to differ in a lot of ways.

### Kstem and possibly stemming in general isn't autocomplete friendly

As we saw with the Brooklyn test, kstem implicitly assumes it is given full words, so it will do strange things n the middle of words that lead to jumpy results

Fixes pelias/schema#89

### DIFF
--- a/integration/analyzer_peliasIndexOneEdgeGram.js
+++ b/integration/analyzer_peliasIndexOneEdgeGram.js
@@ -36,9 +36,9 @@ module.exports.tests.analyze = function(test, common){
     assertAnalysis( 'unique', '1 1 1', ['1'] );
     assertAnalysis( 'notnull', ' / / ', [] );
 
-    assertAnalysis( 'kstem', 'mcdonalds', ['m', 'mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald'] );
-    assertAnalysis( 'kstem', 'McDonald\'s', ['m', 'mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald'] );
-    assertAnalysis( 'kstem', 'peoples', ['p', 'pe', 'peo', 'peop', 'peopl', 'people'] );
+    assertAnalysis( 'no kstem', 'mcdonalds', ['m', 'mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald', 'mcdonalds'] );
+    assertAnalysis( 'no kstem', 'McDonald\'s', ['m', 'mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald', 'mcdonalds'] );
+    assertAnalysis( 'no kstem', 'peoples', ['p', 'pe', 'peo', 'peop', 'peopl', 'people', 'peoples'] );
 
     // remove punctuation (handled by the char_filter)
     assertAnalysis( 'punctuation', punctuation.all.join(''), ['-','-&'] );
@@ -115,7 +115,7 @@ module.exports.tests.functional = function(test, common){
     ]);
 
     assertAnalysis( 'place', 'Toys "R" Us!', [
-      't', 'to', 'toy', 'r', 'u', 'us'
+      't', 'to', 'toy', 'toys', 'r', 'u', 'us'
     ]);
 
     assertAnalysis( 'address', '101 mapzen place', [
--- a/integration/analyzer_peliasIndexTwoEdgeGram.js
+++ b/integration/analyzer_peliasIndexTwoEdgeGram.js
@@ -41,9 +41,9 @@ module.exports.tests.analyze = function(test, common){
     assertAnalysis( 'unique', '11 11 11', ['11'] );
     assertAnalysis( 'notnull', ' / / ', [] );
 
-    assertAnalysis( 'kstem', 'mcdonalds', ['mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald'] );
-    assertAnalysis( 'kstem', 'McDonald\'s', ['mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald'] );
-    assertAnalysis( 'kstem', 'peoples', ['pe', 'peo', 'peop', 'peopl', 'people'] );
+    assertAnalysis( 'no kstem', 'mcdonalds', ['mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald', 'mcdonalds'] );
+    assertAnalysis( 'no kstem', 'McDonald\'s', ['mc', 'mcd', 'mcdo', 'mcdon', 'mcdona', 'mcdonal', 'mcdonald', 'mcdonalds'] );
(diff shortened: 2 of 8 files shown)


orangejulius:
I'm currently importing all of North America from OSM and OA onto my machine to test this, will report back when it's done

orangejulius:
Also, for those who are curious, I tracked down the original [PDF](http://ciir.cs.umass.edu/pubfiles/ir-35.pdf) where the kstem algorithm is described

orangejulius:
Here's a case that may be affected by kstem: [Search for Moscone with focus in SF](http://pelias.github.io/compare/#/v1/search%3Ffocus.point.lon=-122.399505&focus.point.lat=37.784436&text=moscone)

orangejulius:
Hello,
I'm here with an update of running this on dev. I accidentally ran this with code based on master but BEFORE the [autocomplete improvements](https://github.com/pelias/schema/pull/127) were merged in, but this actually helps us see what impact this PR would have a little better, since the autocomplete improvements do fix some issues that would have also been resolved by removing kstem.

Here's the summary overall:

# Acceptance tests (production branch)
For this image and all subsequent ones, production is on the left, and dev is on the right.
screenshot_2016-04-27_13-26-42

So we've fixed two failing cases, but introduced two new regressions.

## First improvement: McDonald's
We previously thought of McDonald's as an example of a case NOT to remove kstemming, but in fact it's the strongest improvement. Right now, if a user enters the text string "mcdonalds", or "mconald", we can't tell the difference. Both are reduced to "mcdonald", so those admin regions named Mcdonald will always win (admin regions get boosts when matching on ngrams because they have 1grams indexed, so venues don't stand a chance against them). Compare our search query for [McDonald's with a focus in NYC](http://pelias.github.io/compare/#/v1/search%3Ffocus.point.lat=40.743987&focus.point.lon=-73.990417&text=mcdonalds)
screenshot_2016-04-27_13-35-13
Perfect clustering :tada: 

This also is evident in a query for [Moscone center with a focus in sf](http://pelias.github.io/compare/#/v1/search%3Ffocus.point.lon=-122.399505&focus.point.lat=37.784436&text=moscone). Previously we could not distinguish Moscone from Moscones, which is a city in Peru. There are also admin regions called moscone, so this search isn't quite as clear, but you can see the expected Moscone Center related results have moved up.

## Second improvement: Brooklyn autocomplete (and Daly City)

This is the same as in the PR body above. `brookly` is reduced by kstem to `brook`, which is then shortened. Note that while the recent autocomplete changes fix the abbreviation issue, which means `brook` is no longer replaced with `brk`, results that match `brook` are still shown on prodbuild in a [partial autocomplete query for `brookly`](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=brookly) while all results in dev match brookly. Prod has neither and shows things matching `brk` which is not much.
screenshot_2016-04-27_13-39-44

Likewise for [Daly city](http://pelias.github.io/compare/#/v1/autocomplete%3Ftext=daly%20city). Kstem changes `daly` to `dale`, and while the autocomplete improvements help, matches for `dale` still show up as high as 2nd. On dev, there are no `dale`-related results, only `daly` results!
screenshot_2016-04-27_13-42-58

## Regressions: [1 water st manhattan ny](http://pelias.github.io/compare/#/v1/search%3Ftext=1%20water%20st%20manhattan%20ny)
This doesn't seem to be kstem related, and I believe improving our admin matching with [multi_match](https://github.com/pelias/api/pull/481) and similar work will fix this.

# Fuzzy tests
Overall there's a dramatic improvement, with over 30 queries passing where they weren't before
screenshot_2016-04-27_13-17-18

Some of the improvements are similar to above so I'll only touch on them briefly
## brooklyn
screenshot_2016-04-27_13-10-59

## Mcdonalds
screenshot_2016-04-27_13-15-01

One larger improvement is the Mcdonalds query in Singapore:
## McDonald's, Central Singapore, Singapore
screenshot_2016-04-27_13-10-29
Note that because only results matching `mcdonalds`, not `mcdonald` are returned, the correct result it shown MUCH sooner than before! :palm_tree:

A fix that I haven't fully looked into comes from the new WOF tests:
## [partial search for Comunidad Foral De Navarra, ES](http://pelias.github.io/compare/#/v1/autocomplete%3Fsources=wof&text=Comunidad%20Fora)

I can't find any difference in the way `fora` is organized, but in any case, we were not returning any results here before, which was causing jitter. Now the jitter is gone.


So, there are lots of improvements. I'm going to kick off another dev build soon which rebases the kstem removal on top of the recent autocomplete changes, so we can see how the two combine. I predict the results will be even better, and we should finally merge this!

missinglink:
:+1: LGTM

orangejulius:
:scream: it got a :+1: !!!!

PR #95, opened 2016-01-12
use WOF placetypes instead of quattroshapes placetypes

missinglink:
**[edit]** this PR has been updated in order to support both quattroshapes and WOF *simultaneously* and should now be good-to-release.

this PR changes the terminology used to refer to the 'administrative' hierarchy of geographic regions which parent a document.

prior to this PR these document fields were named:

```
admin0
admin1
admin1_abbr
admin2
local_admin
locality
neighborhood
```

the proposal in this PR is to **add** the following fields:

```
parent.country
parent.country_abbr
parent.country_id
parent.region
parent.region_abbr
parent.region_id
parent.county
parent.county_abbr
parent.county_id
parent.locality
parent.locality_abbr
parent.locality_id
parent.localadmin
parent.localadmin_abbr
parent.localadmin_id
parent.neighbourhood
parent.neighbourhood_abbr
parent.neighbourhood_id
```

I chose to store the fields inside an 'object' type (kind of like a namespace) called `"parent"`, I personally think this is 'nicer' simply because it's easier to distinguish these fields from the other fields in the document.

The `alpha3` field remains unchanged.

I added 3x fields per place type, the default being the 'name' and the other 2 being `id` and `abbr`. In the past only `admin1` allowed for state abbreviations.

One further consideration is the names of the elasticsearch `_type` that we tell to use a single gram analysis. Prior to this PR we created 3x `_type` at index time named `admin0, admin1, admin2` which were set up to index single grams (so the first keypress returns something).

These `_type` are now named `country, region, county` respectively. It's important that when importing WOF placetypes that the `_type` they get stored in is named the same, failing to do so will return 0 results for the first keypress.

For more info about this see my comment https://github.com/pelias/schema/blob/admin_field_naming/schema.js, the simple solution is to use the WOF placetype as the first argument to the pelias/model constructor such as:
```javascript
// good
var poi = new Document( 'wof', 'country', 1 )

// bad
var poi = new Document( 'wof', 'wof:country', 1 )
```

I also took the opportunity to do a little housekeeping and add some code comments.

### DIFF
--- a/integration/address_matching.js
+++ b/integration/address_matching.js
@@ -8,8 +8,8 @@ var tape = require('tape'),
 
 module.exports.tests = {};
 
-module.exports.tests.source_filter = function(test, common){
-  test( 'mapzen hq', function(t){
+module.exports.tests.functional = function(test, common){
+  test( 'functional', function(t){
 
     var suite = new elastictest.Suite( null, { schema: schema } );
     suite.action( function( done ){ setTimeout( done, 500 ); }); // wait for es to bring some shards up
--- a/integration/admin_matching.js
+++ b/integration/admin_matching.js
@@ -1,0 +1,113 @@
+
+// validate analyzer is behaving as expected
+
+var tape = require('tape'),
+    elastictest = require('elastictest'),
+    schema = require('../schema'),
+    punctuation = require('../punctuation');
+
+module.exports.tests = {};
+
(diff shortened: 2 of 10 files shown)


dianashk:
Anything still left to be done here? Looks good to me.

trescube:
Merged

Issue #94, opened 2016-01-07
possible regex issue


missinglink:
The `remove_ordinals` regex seems to be wrong?
```
        "remove_ordinals" : {
          "type" : "pattern_replace",
          "pattern": "(([0-9])(st|nd|rd|th))",
          "replacement": "$2"
        },
```

this should really be `(([0-9]+)(st|nd|rd|th))` (with a plus sign) shouldn't it?

the weird thing is the test is passing here: https://github.com/pelias/schema/blob/master/integration/analyzer_peliasStreet.js#L27

ref: https://github.com/pelias/schema/blob/master/settings.js#L151

orangejulius:
We fixed this in #96! :cherries:

PR #96, opened 2016-01-12
better remove_ordinals regex pattern

missinglink:
see https://github.com/pelias/schema/issues/94 for more info.

@orangejulius could you review this plz?

closes https://github.com/pelias/schema/issues/94

[edit] so after much yak shaving I have come up with the perfect megaregex `replace( "(?i)((^| )((1)st?|(2)nd?|(3)rd?|([4-9])th?)|(([0-9]*)(1[0-9])th?)|(([0-9]*[02-9])((1)st?|(2)nd?|(3)rd?|([04-9])th?))($| ))", "$2$4$5$6$7$9$10$12$14$15$16$17$18"`

...which makes my eyes hurt but makes all the tests pass.

@trescube suggested we should maybe leave it a bit more relaxed so that users who are not perfect at english can still find what they need.

so there are 2 options, either we go with the super complex one above or we modify the existing one to support autocomplete with the simpler: `replace( "(([0-9])(st?|nd?|rd?|th?))", "$2" )` or similar

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -148,8 +148,8 @@ function generate(){
         },
         "remove_ordinals" : {
           "type" : "pattern_replace",
-          "pattern": "(([0-9]*1)st?|([0-9]*2)nd?|([0-9]*3)rd?|([0-9]*[456789])th?|([0-9]+0)th?|([0-9]*1[0-9])th?)",
-          "replacement": "$2$3$4$5$6$7"
+          "pattern": "(?i)((^| )((1)st?|(2)nd?|(3)rd?|([4-9])th?)|(([0-9]*)(1[0-9])th?)|(([0-9]*[02-9])((1)st?|(2)nd?|(3)rd?|([04-9])th?))($| ))",
+          "replacement": "$2$4$5$6$7$9$10$12$14$15$16$17$18"
         },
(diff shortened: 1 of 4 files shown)


settings.js
```diff
@@ -148,8 +148,8 @@ function generate(){
         },
         "remove_ordinals" : {
           "type" : "pattern_replace",
-          "pattern": "(([0-9])(st|nd|rd|th))",
-          "replacement": "$2"
+          "pattern": "(([0-9]*1)st?|([0-9]*2)nd?|([0-9]*3)rd?|([0-9]*[456789])th?|([0-9]+0)th?|([0-9]*1[0-9])th?)",
```
orangejulius: :mind_blown: This is quite the regex. Looks like it does what you want though

orangejulius:
This looks like an improvement over the old regex for sure. I'd be very interested to do some testing of this on dev. 

Somewhat related: have we tested not removing ordinals? There's definitely a big signal difference between 5th and 5. Just like in #93 it's not possible to find places named "whole food market" with kstemming because places named "whole foods" evaluate to the same tokens, having 5 and 5th indistinguishable seems less than ideal.

PR #101, opened 2016-02-24
Experimental Elasticsearch 2.x support

### LINKED ISSUE #106: Create index script fails on newer elasticsearch v2.2
I tried running it against latest elasticsearch v2.2 and it fails with error :

[put mapping] 	 pelias { [Error: [mapper_parsing_exception] analyzer on field [neighbourhood_id] must be set when search_analyzer is set]
  status: '400',
  message: '[mapper_parsing_exception] analyzer on field [neighbourhood_id] must be set when search_analyzer is set',
  path: '/pelias',
  query: {},
...

orangejulius:
After Elasticon last week I spent an hour or so trying out Elasticsearch 2.2 on my machine. There were only a few changes to be made, and for the most part they are fairly trivial.

There is one exception: as detailed in [3fa56ef](https://github.com/pelias/schema/commit/3fa56ef695705bec1563ef749fd6d0680f6f4738), we can't use different mappings on fields of the same name in different types. This is extremely unfortunate for us, since we currently use the peliasOneEdgeGram analyzer in mappings for city, country, region, and other admin fields where we want autocomplete to perform really well, but the peliasTwoEdgeGram analyzer in mappings for all other _types.

We can't do this anymore so I've made this PR switch everythithing to peliasOneEdgeGram. This might increase our index size a bit and more importantly, will reduce query performance since the number of documents for each of the 1 character n-grams will be huge, and Elasticsearch will have to score all of them.

Before this is merged I intend to do some analysis of index size and search performance. We probably want to take another look at the [completion suggester](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html) again, since in 2.x it supports better geo flitering than 1.x.

Fixes #106

### DIFF
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -111,7 +111,6 @@ var schema = {
         type: 'string',
         analyzer: 'peliasOneEdgeGram',
         fielddata : {
-          format : 'fst',
           loading: 'eager_global_ordinals'
         }
       }
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -24,23 +24,19 @@ var schema = {
       properties: {
         name: {
           type: 'string',
-          index_analyzer: 'keyword',
-          search_analyzer: 'keyword'
+          analyzer: 'keyword'
         },
         number: {
           type: 'string',
-          index_analyzer: 'peliasHousenumber',
-          search_analyzer: 'peliasHousenumber',
+          analyzer: 'peliasHousenumber',
         },
         street: {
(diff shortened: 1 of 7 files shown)


missinglink:
interesting experiment :+1: for giving a whirl, if performance is poor we can talk options.

orangejulius:
Who needs this branch when we have _REAL_ Elasticsearch 2 support? ( #138 )

PR #107, opened 2016-03-11
add address_suffix_expansion to peliasOneEdgeGram analysis, add tests

missinglink:
pretty straight forward, this PR adds the existing `address_suffix_expansion` filter to the `peliasOneEdgeGram` analyzer.

the motivation comes from using the `peliasOneEdgeGram` as our default search analyzer so it's best if it works well when used against and index created using `peliasTwoEdgeGram`


orangejulius:
Makes sense :+1:

PR #111, opened 2016-03-23
Index one-grams for macroregion, macrocounty, localadmin, locality

orangejulius:
This will prevent localities and other admin areas from ranking below
counties that have the same name, but lower population.

This hopefully fixes the failing San Francisco acceptance test found in https://github.com/pelias/acceptance-tests/issues/192#issuecomment-199464239

### DIFF
--- a/schema.js
+++ b/schema.js
@@ -27,21 +27,25 @@ var schema = {
     _default_: doc,
 
     /**
-      these 3 _type are created when the index is created, while all other _type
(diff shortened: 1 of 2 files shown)


trescube:
:+1:

orangejulius:
A build with this change hasn't fully finished on dev, but things appear to be working well so far. San Francisco, CA and Las Vegas, NV both come up first in searches now.

PR #125, opened 2016-04-15
Warn when dropping the schema and not connected to localhost

orangejulius:
:sob:


trescube:
YES YES YES!  🚢 🐏 👍 ⚡

trescube:
I appreciate things that point the shotgun away from my foot.

PR #138, opened 2016-06-30
Elasticsearch2

missinglink:
- add `goldbergianPeliasHousenumber` to ensure that numeric values are treated as whole tokens rather than creating prefixes.
- use single gram analyzers for all fields, no longer using 2 gram analysis
- remove settings deprecated in ES2
- remove `dynamic_templates` so now all `_type` have the same config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ env:
    - CXX=g++-4.8
  matrix:
    - TEST_SUITE=test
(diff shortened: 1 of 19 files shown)


orangejulius:
:shipit:

dianashk:
:shipit:

avulfson17:
LGTM

Issue #74, opened 2015-10-01
expand umlaut, eg. 'ö' -> 'oe'


missinglink:
it's common for users speaking germanic languages to expand `ö` to `oe`, especially when stuck using an English keyboard.

an example of this is `Löhningen` which we can currently surface using `Löhningen` and the `asciifolded` version `Lohningen` but not by the term `Loehningen`.

other vowels are also effected, they are called `"umlaut"` in German, there may be an existing Lucene analyser for this which we can use.

ref: https://en.wikipedia.org/wiki/%C3%96
thanks @ralphiech for reporting

ralphiech:
This essentially is relevant for the following chars

```ü ö ä Ä Ü Ö```

orangejulius:
Like https://github.com/pelias/schema/issues/65#issuecomment-237240507, this is now possible thanks to our alt-name support. It's something we should look at again.

PR #146, opened 2016-07-21
expand umlauts

missinglink:
[DO NOT MERGE] at this time

``` javascript
"umlaut" : {
  "type" : "mapping",
  "mappings" : [
    "ä=>ae",
    "ö=>oe",
    "ü=>ue",
    "Ä=>Ae",
    "Ö=>Oe",
    "Ü=>Ue",
    "ß=>ss"
  ]
}
```

closes https://github.com/pelias/schema/issues/74
closes https://github.com/pelias/pelias/issues/319

### DIFF
--- a/integration/analyzer_peliasAdmin.js
+++ b/integration/analyzer_peliasAdmin.js
@@ -15,6 +15,15 @@ module.exports.tests.analyze = function(test, common){
     var assertAnalysis = analyze.bind( null, suite, t, 'peliasAdmin' );
     suite.action( function( done ){ setTimeout( done, 500 ); }); // wait for es to bring some shards up
 
+    // expand umlauts
+    assertAnalysis( 'umlaut', 'Häuser', ['haeuser']);
+    assertAnalysis( 'umlaut', 'Malmö', ['malmoe']);
+    assertAnalysis( 'umlaut', 'Bücher', ['buecher']);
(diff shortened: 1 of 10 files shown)


missinglink:
Fun fact of the day:

In Germany, A Big Mac used to be a Big Mäc.

And the Filet-o-Fish was the Fishmäc. The spelling with the umlaut actually gets German speakers a little closer to the English pronunciation of “Mac.” But in 2007 McDonald’s took away the umlauts, and now Germans have to get boring ol’ Big Macs like the rest of us.

pulp

missinglink:
This PR would break some existing behavior, eg:

records indexed as `Österreich` would be query-able using the input `Oesterreich` but not `Osterreich`.

prior to this PR the term `Österreich` was retrievable using the term `Osterreich` so this would be a breaking change.

It seems like a better strategy would be to import two versions of the same token; once as `Oesterreich` and once as `Osterreich`.

orangejulius:
Do we want to close this PR in favor of something icu related?

missinglink:
I don't believe the ICU library will handle this type of expansion.

missinglink:
This feature will require alt-names as the country name above can have 3 forms:

```
Österreich
Oesterreich
Osterreich
```

moving to alt-names feature as it can only be solved for a maximum of 2 cases before then.

trescube:
Closed due to teh new [hotness](pelias/placeholder)

Issue #148, opened 2016-07-26
use icu_normalizer


missinglink:
prior to elasticsearch2 the [icu_normalizer](https://www.elastic.co/guide/en/elasticsearch/guide/current/unicode-normalization.html) was only available as a plugin so we didn't include it, it seems to now be included in the core engine.

we should enable this filter for all analysis in order to properly handle unicode combining characters.

missinglink:
see https://github.com/pelias/api/issues/600 for more info and test case

missinglink:
It turns out that although the ICU plugin is now shipped with the elasticsearch installer, it is not enabled by default :(

see: 
- https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu.html
- https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis.html

I would be :+1: fo enabling it however it would require every user to execute the follow command during installation/setup:

```
sudo bin/plugin install analysis-icu
```

Failure to do so would result in a fatal error message such as `"Unknown TokenFilter type [icu_normalizer] for [nfkc_normalizer]"` and the Pelias installation would be unusable.

missinglink:
The work here is to submit a PR with a DO NOT MERGE label, then discuss the issues around installing the plugin.

missinglink:
PR https://github.com/pelias/schema/pull/149 created to update schema, still pending a discussion on how best to document the plugin installation process.

missinglink:
blocked by ops ticket to enable for *mapzen search* https://github.com/mapzen/operations-engineering/issues/346 (private repo)

PR #149, opened 2016-07-27
refactor schema to use icu plugin

missinglink:
DO NOT MERGE at this time, this PR requires that the `icu plugin` to be installed in elasticsearch in order to work.

Closes https://github.com/pelias/schema/issues/148

More info can be found on the issue above, this PR enables the ICU plugin and uses unicode-aware functions in place of non unicode-aware functions.

In effect this provides support for unicode combining characters and better glyph folding.

### DIFF
--- a/integration/analyzer_peliasAdmin.js
+++ b/integration/analyzer_peliasAdmin.js
@@ -99,6 +99,38 @@ module.exports.tests.tokenizer = function(test, common){
   });
 };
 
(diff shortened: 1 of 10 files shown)


orangejulius:
We kicked off a build on Dev2 last night, and this seems to have worked fine, so LGTM. When should we merge it, are we still waiting for anything?

dianashk:
:+1:

PR #153, opened 2016-07-28
Pass config to Elasticsearch

orangejulius:
One confusing part of `esclient` is that it automatically pulls in `pelias-config`. I know this because I missed the exact same thing when making this change for the API: https://github.com/pelias/api/pull/448

### DIFF
--- a/scripts/create_index.js
+++ b/scripts/create_index.js
@@ -1,5 +1,6 @@
+var config = require('pelias-config').generate().esclient;
 var es = require('elasticsearch');
-var client = new es.Client();
+var client = new es.Client(config);
 var schema = require('../schema');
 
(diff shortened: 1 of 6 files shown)


trescube:
:shipit:

scripts/create_index.js
```diff
@@ -1,5 +1,6 @@
+var config = require('pelias-config').generate();
```
dianashk: should this also say `generate().esclient` like all the other files below?

orangejulius: Indeed it should. I also missed one whole file. Thanks!

dianashk:
Yay! LGTM

orangejulius:
Oh there was one more tiny change, one file already used `pelias-config`

PR #167, opened 2016-08-18
added `dependency` in the various places

trescube:
`dependency` was missing from the various analysis and search pieces.


dianashk:
:+1: trying to bring back the thumbs

orangejulius:
LGTM

PR #170, opened 2016-09-06
Add locale support into street suffixes

valotvince:
Hi, 

We have been working on a POC for a customer using Pelias with specific needs, especially on the area of Paris in France. We only populated es with the data from Paris, with french street suffixes.

With the current Pelias schema, we aren't able to type, for exemple, "bd" which stands for "Boulevard".

I have been working on a little feature that allows to add specific locale terms and suffixes into the Pelias schema (default is only en). Working well for now on our dev env. The other locales could be added into the "locales" directory and loaded only if the dev wants to do so (adding a key "locales" into "schema" object of Pelias config).

Shipping the locales with Pelias schema could be an efficient way to get a standard street_suffix file for each locale, instead of loading it on our side.

Please let me know if it really fits with Pelias and, if not, what I could improve :)

Thanks,
Vincent.


valotvince:
An another way is to load "en" locale everytime and then specify extra_locales to add.

missinglink:
I like this a lot! It will be great to have per-locale settings in Pelias and I am :+1: for adding them to the source.

I have one concern though; which is how they are configured.

This implementation allows developers to have a :fr: pelias or an :gb: pelias but does not allow for both.

I think an ideal solution would be to partition data by language and apply these local settings to fields depending on their language, this would allow us to create a :globe_with_meridians: pelias which works for everyone.

Re: merging this PR, I am 100% :+1: on merging the `locales/street_suffix.fr.js` file.

I am also ok with merging the changes to `generate()` so long as people understand that we will likely remove the all-or-nothing locale setting in the future.

Interested to hear other peoples opinions, great work @valotvince thanks for the PR.

missinglink:
note: merging this PR will require another PR to add `config.schema.locales : ["en"]` to `config/defaults.json` in https://github.com/pelias/config

missinglink:
FYI: https://github.com/openvenues/libpostal/tree/master/resources/dictionaries/fr

valotvince:
Hi missinglink,

> This implementation allows developers to have a :fr: pelias or an :gb: pelias but does not allow for both.
I think an ideal solution would be to partition data by language and apply these locale settings to fields depending on their language, this would allow us to create a :globe_with_meridians: pelias which works for everyone.

This implementation does allow only french, only english (that's how we have configured our schema), both of them or many more.

> note: merging this PR will require another PR to add config.schema.locales : ["en"] to config/defaults.json in https://github.com/pelias/config

For pelias-config, I was thinking of it during lunch haha. For now, I check if the locales are defined before using them, and defaulting to "en" only.

> FYI: https://github.com/openvenues/libpostal/tree/master/resources/dictionaries/fr

I didn't find those on Google, I was using a part of one definition of our geocoder :) I will change the street_suffix.fr.json file according to it.

I'm glad you enjoyed it :)

Vincent.

trescube:
We'll move forward with this in 2017 Q3 to allow for multiple languages now that we're all in on libpostal.  We'll also incorporate data from [MapBox's data](https://github.com/mapbox/geocoder-abbreviations/tree/master/tokens).

missinglink:
Hi @valotvince,

Pelias is still very an English focussed geocoder and [we're taking steps towards better internationalization](https://mapzen.com/blog/geocoding-for-polyglots/).

Some work has been done recently in [pelias/analysis](https://github.com/pelias/analysis) to improve support for per-locale text analysis, including synonym substitution.

In the long-term, I think there is much more flexibility in handling the analysis in javascript rather than in elasticsearch, this will allow us to run a geocoder that speaks all languages rather than only one.

I've opened an issue on [that repo](https://github.com/pelias/analysis/issues/4) to ensure that your work is merged in to pelias/analysis.

Thank you for your contribution, I'm sorry we weren't able to merge this PR.

PR #172, opened 2016-09-09
Move elasticsearch to dependencies

orangejulius:
It's required in production but has been living in devDependencies for a
long time since we were installing both.


dianashk:
oooooh good one!

:shipit:

PR #174, opened 2016-09-16
Remove space appending to directionals

trescube:
The reason that `186 Tuskegee St SE Atlanta GA` test was failing is because directionals were being contracted with a space appended which is incorrect behavior since postdirectionals are a thing.


missinglink:
:shipit: looks good to me

dianashk:
Looks like LGTM doesn't recognize the fancy new github review comments. So :+1:

Issue #190, opened 2016-11-14
Remove useage of relative paths from scripts


orangejulius:
Some of our scripts, most notably [create_index.js](https://github.com/pelias/schema/blob/master/scripts/create_index.js#L10), use relative paths when loading other Javascript modules. This causes them to fail if not run from the root directory of this repository.

For example:
```bash
git clone git@github.com:pelias/schema.git
cd schema
npm install

# works
node scripts/create_index.js

# breaks
cd scripts; node create_index.js
```

PR #219, opened 2017-04-06
Fixed path to check_plugins.js in create_index.js

eugenezinoviev:
Fixed path to check_plugins.js in create_index.js script to allow run index creation from any working dir only only from schema directory itself.

Fixes https://github.com/pelias/schema/issues/190


orangejulius:
Thanks @eugenezinoviev!

PR #226, opened 2017-06-04
Update check_plugins.js

bradh:
Reflect new name for plugin tool (as of Elasticsearch 5.0)


bradh:
Ah, my bad. Missed that it has to be ES 2.3.

orangejulius:
Hey @bradh,
So if I'm understanding correctly, you're saying the name of the plugin tool changed in Elasticsearch 5? Can you link to any info you have about that? It's definitely something we'll want to keep in mind for the eventual transition to ES5. Thanks!

bradh:
@orangejulius I actually tried against ES 5, and that is what it was called. The docs show the difference (as of ES 5.0). Compare https://www.elastic.co/guide/en/elasticsearch/plugins/2.4/plugin-management.html and https://www.elastic.co/guide/en/elasticsearch/plugins/5.0/plugin-management.html as an example

PR #227, opened 2017-06-13
Rename "Scripts" section to "Usage"

tigerlily-he:
All that is required for setting up the schema is downloading the node modules and creating the index

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,6 @@ $ npm install pelias-schema
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


orangejulius:
On second thought, I'm not sure this actually needs to be deleted. But it certainly isn't the most important thing in the readme and should not be first. The current "scripts" section should be renamed to usage. If we think it's worth keeping around, we can move the section on how to print out the schema much lower in the readme.

PR #234, opened 2017-08-21
'completed token' synonyms

missinglink:
[DO NOT MERGE]
NOTE: based on https://github.com/pelias/schema/pull/237, please merge that PR first

fixes https://github.com/pelias/pelias/issues/563

we maintain two types of synonym lists:

- synonyms safe for 'partial tokens' eg. 'rd' **but not** 'st'
- synonyms safe for 'complete tokens' eg. **both** 'rd' and 'st'

this PR fixes a bug where the `full_token_address_suffix_expansion` filter was only using the 'partial safe' synonyms (meaning 'st' => 'street' was not available).

as a result I have reconfigured the `partial_token_address_suffix_expansion` to behave as the `full_token_address_suffix_expansion` used to, and migrated the `peliasIndexOneEdgeGram` and `peliasIndexTwoEdgeGram` analyzers across to using `partial_token_address_suffix_expansion` instead.

the tl;dr is that searches for complete tokens should have improved synonyms substitution while searches for partial tokens should remain unchanged.

note: this may also fix some open issues we have regarding synonym substitution for `/search` and `/search/structured`, I'll dig through the backlog and see what I can find.


orangejulius:
A build on dev with these changes had been kicked off!

missinglink:
I'm going to abandon this PR, as I remembered more about how we currently do things, I became more convinced that this PR would result in worse behaviour rather than better.

PR #238, opened 2017-08-28
update schema mappings to include all current layers

missinglink:
the schema allows creating new `_types` dynamically using the `_default_` API, we have been using this feature when introducing new layers such as `street`.

there is, however, an advantage to creating the `_type` explicitly. Elasticsearch will error instead of returning 0 results for any `_type` that is requested but is not present in the `_mapping`.

by creating all known layers at schema creating time, we can avoid these sort of errors, this is less relevant for full-planet, all-layer builds and more relevant to smaller builds which only import a subset of layers.

I haven't seen a bug report for it, but I imagine that if someone *only* imported WOF and requested something from a combination of `street` or `region` layers, a 500 level error would be produced rather than simply returning matches from the `region` layer.

### DIFF
--- a/schema.js
+++ b/schema.js
@@ -19,15 +19,20 @@ var schema = {
 
       querying against non-existant _types will result in errors.
     **/
-    country: doc,
-    dependency: doc,
-    macroregion: doc,
-    region: doc,
-    macrocounty: doc,
-    county: doc,
-    localadmin: doc,
+    venue: doc,
+    address: doc,
+    street: doc,
+    neighbourhood: doc,
+    borough: doc,
+    postalcode: doc,
     locality: doc,
-    borough: doc
+    localadmin: doc,
+    county: doc,
+    macrocounty: doc,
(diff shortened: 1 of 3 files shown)


missinglink:
my understanding is that a `type` was introduced to give the feeling of a `table` (ie. a logical subset of a `database`) but internally all the data is actually all sitting in the same index, it is simply implemented as an indexed-filter operation on the index.

if that's the case then we could begin to introduce a `layer` field which is a copy of the `_type` and then migrate the queries over to use a boolean filter instead of the existing. `type` API.

doesn't sound technically difficult but sounds time-consuming and error-prone.

orangejulius:
So I knew something about this seemed familiar. It turns out we've been filtering on a layer field for [over a year](https://github.com/pelias/api/pull/580). This means that as we desire, we can start to remove usage of `_type` all together, and definitely in time for Elasticsearch changes which are still a long way out.

PR #258, opened 2017-11-02
Adding address unit attribute to schema

sweco-semhul:
To resolve https://github.com/pelias/pelias/issues/618 more details found in comments to that issue

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -117,6 +117,15 @@ function generate(){
             "trim"
           ]
         },
+        "peliasUnit": {
+          "type": "custom",
+          "tokenizer":"keyword",
+          "char_filter" : ["alphanumeric"],
+          "filter": [
+            "lowercase",
+            "trim"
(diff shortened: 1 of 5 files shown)


settings.js
```diff
@@ -117,6 +117,15 @@ function generate(){
             "trim"
           ]
         },
+        "peliasUnit": {
+          "type": "custom",
+          "tokenizer":"keyword",
+          "char_filter" : ["alphanumeric"],
```
orangejulius: @trescube do you think we would ever need non-alphanumeric characters in unit numbers? Maybe `-` or `/`?

sweco-semhul:
yay thanks!

settings.js
orangejulius: We've had a discussion and are pretty sure analyzing alphanumeric characters only will work well enough to start :)

PR #266, opened 2018-01-30
Replace pelias-mergeable with lodash.merge

Echelon9:
[pelias-mergeable](https://github.com/pelias/mergeable/) is to be deprecated, so replace usage with the functionality provided by `lodash`.

Given we only require a small subset of lodash, add dependency for `lodash.merge` to avoid pulling in an unnecessarily large dependency.

See further the objective in: https://github.com/pelias/mergeable/issues/5

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "colors": "^1.1.2",
     "elasticsearch": "^13.3.1",
     "joi": "^12.0.0",
-    "mergeable": "latest",
+    "lodash.merge": "^4.6.0",
     "pelias-config": "2.13.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "colors": "^1.1.2",
     "elasticsearch": "^13.3.1",
     "joi": "^12.0.0",
-    "mergeable": "latest",
+    "lodash.merge": "latest",
```
orangejulius: Can you change this to whatever the latest version is? That way Greenkeeper will help us detect breaking changes.

Echelon9: Sure, I'll update the PR to reflect that.

Echelon9:
@orangejulius done! Ready to merge.

PR #278, opened 2018-04-15
Dockerfile tweaks

orangejulius:
Speed up repeated builds of code-only changes by copying package.json and running npm install separately first.

Additionally, disable unit test running as part of the build process. There are many cases where we might want to build docker images that don't have passing unit tests.


orangejulius:
Merging sans review to build on top of this. I kinda do not feel great about disabling the unit tests in general.

PR #282, opened 2018-05-01
Use strict mapping

orangejulius:
Strict [dynamic mapping](https://www.elastic.co/guide/en/elasticsearch/guide/current/dynamic-mapping.html) tells Elasticsearch to throw an exception whenever an attempt to index a field not specified in the type mapping is made.

By setting dynamic mapping to true, we missed issues such as [continents being added to the model, but not schema](https://github.com/pelias/schema/pull/281). Using strict mapping should help us catch issues like this in the future.

### DIFF
--- a/integration/admin_matching.js
+++ b/integration/admin_matching.js
@@ -17,26 +17,28 @@ module.exports.tests.functional = function(test, common){
     suite.action( function( done ){
       suite.client.index({
         index: suite.props.index, type: 'test',
-        id: '1', body: { admin: {
-          country: 'Test Country',
-          country_a: 'TestCountry',
-          country_id: '100',
-          region: 'Test Region',
-          region_a: 'TestRegion',
-          region_id: '200',
-          county: 'Test County',
-          county_a: 'TestCounty',
-          county_id: '300',
-          locality: 'Test Locality',
-          locality_a: 'TestLocality',
-          locality_id: '400',
-          localadmin: 'Test LocalAdmin',
-          localadmin_a: 'TestLocalAdmin',
-          localadmin_id: '500',
-          neighbourhood: 'Test Neighbourhood',
-          neighbourhood_a: 'TestNeighbourhood',
-          neighbourhood_id: '600',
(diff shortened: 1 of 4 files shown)


missinglink:
Is strict mapping compatible with dynamic_templates?

orangejulius:
It appears to be, yes. You can set it at each level so, for example, at the top level of our type mapping, strict mapping is enabled. But within the name and phrase objects, there is a dynamic template.

I have run Portland imports with this branch, and all tests seemed to be fine. :)

orangejulius:
I'd still like to consider merging this. True, it would require anyone doing anything custom to ensure their schema was in line with their custom changes, but that's a good idea.

Elasticsearch can have significant performance and disk usages problems if fields aren't indexed correctly, and we have many times experienced that after changing our admin fields. It would be nice to be able to avoid it.

An alternative to this pull request would be something like this:

We separate the parent id, parent name, and parent abbreviation fields. Instead of all living under the `parent` object, there would be `parent_id`, `parent_name` and `parent_abbr` fields at the root level of our documents.

Each of those root level fields would have a dynamic template ensuring any sub-field has the correct mapping for that type of data.

This would be a pretty big refactoring of our schema, and would require code updates all over the place, but it would make it really easy to add new parent fields while ensuring they have the correct mappings. That could in many ways be more valuable as we look to enable more customization in Pelias.

missinglink:
I would be interested in hearing from the community on this issue, it doesn't really affect us as we control the standard schema.
I suspect this might have a large impact on those running non-standard schemas, if not then I'm :+1: to merge it.

Any comments from the community?
cc/ @pelias/contributors

Joxit:
At Jawg we are using standard pelias schema, so it's ok for us :smile:

orangejulius:
While refreshing this PR, I discovered that one of our integration tests had been using incorrect field names for a long long time. This is exactly the benefit that strict elasticsearch schemas give us, the maintainer! 🎉

PR #291, opened 2018-05-16
Portland synonyms

missinglink:
@orangejulius we need to decide what to do with this code, I would suggest we merge it, but I'm not sure what the effect would be, in particular the following changes make me nervous:

- removing space from the peliasStreetTokenizer
- changing the analyzer token filters
- changing the synonyms files, making some 'real synonyms' instead of replacements and updating the lists.

despite the changes, we should try to get this published, it was tested on a smaller extract and had a positive impact.

how would you like to handle merging, testing and deploying this?

[minor] looks like `RUN npm test` was removed from the dockerfile and the format was changed

### DIFF
--- a/synonyms/custom_name.txt
+++ b/synonyms/custom_name.txt
@@ -23,7 +23,9 @@
 #   foo => foo bar, baz
 #
 # =============================================================================
-mlk,martin luther king junior
-ml,martin luther
-tvh,tualatin valley highway
-tv,tualatin valley
+mlk => martin luther king junior
+ml => martin luther
+ml king=> martin luther king junior
+tvh => tualatin valley hwy
+sw tvh=>sw tualatin valley hwy
+tv => tualatin valley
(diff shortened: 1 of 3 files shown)


orangejulius:
I agree, we should definitely merge the code changes (not the synonyms which are Portland specific).

I think with https://github.com/pelias/schema/tree/disable_fielddata, which I have tested using https://github.com/orangejulius/ci-test and appears to perform identically for Portland, we should be able to now run fairly large builds on our local machines. Want to plan to merge these code changes after comparing the results on a North America or Europe build?

orangejulius:
Thanks to https://github.com/pelias/docker/pull/23 I can now confirm that this PR and https://github.com/pelias/schema/pull/310 perform identically when used in the `pelias/docker` Portland Metro project. We should look towards merging that PR and closing this one when we can.

orangejulius:
All generally relevant code for this PR has now been merged in #310, so this PR is no longer needed.

PR #293, opened 2018-05-18
Use a single Elasticsearch Mapping Type

orangejulius:
We currently set up many Elasticsearch [mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/mapping.html#mapping-type). They are all identical, and thus there's no benefit to having more than one.

In Elasticsearch 6, multiple mapping types will [not be supported](https://www.elastic.co/guide/en/elasticsearch/reference/6.2/removal-of-types.html), so we should get on top of things and remove them when we can.

There's also the immediate benefit that our Elasticsearch mapping will be much shorter and easier to work with. This not only makes Elasticsearch itself have to do less work to keep track of our types, but take a look at the **5000 lines** of duplicated mapping type configuration we get to remove in our tests below!

This PR is built on https://github.com/pelias/schema/pull/282 and must be merged after https://github.com/pelias/model/pull/95

Connects https://github.com/pelias/pelias/issues/719


missinglink:
I don't see any reason we can't merge this, we just need to first check:

- no queries are relying on this functionality (and consider the effect on queries written in forks)
- we fun a full build and test that the quality and performance hasnt changed

The same goes for the strict mapping, but in that case I'm not sure it would provide any real benefit? I'll add my comments on that PR.

orangejulius:
I just updated this PR. When it was originally written, Travis didn't run the integration tests, which now pass.

There's now **7000** lines of duplicate mapping configuration that will be removed :)

It's good to merge at any time, however we should probably wait a month or so after both https://github.com/pelias/model/pull/95 and https://github.com/pelias/model/pull/121 are merged first.

orangejulius:
https://github.com/pelias/model/pull/95 was merged a week ago. I think we should merge this around October 15th!

orangejulius:
Merged this a bit earlier than planned: it will be nice to have ES6 compatible schemas and Elasticsearch builds being completed ASAP, and anyone who runs into compatibility issues will only do so as they are about to start a new build. They will be able to update their importers and be ok.

ES6 here we come!

Issue #100, opened 2016-02-19
Investigate disabling fielddata for name.* fields


orangejulius:
My understanding from reading through the [fielddata docs](https://www.elastic.co/guide/en/elasticsearch/guide/current/fielddata.html) and from the AMA desk at Elasticon is that fielddata is only used for aggregations, sorting, and geo operations on geo fields.

Of those 3, as far as I know, we only do geo operations. However, fielddata is currently enabled on the name.\* fields and in dev is currently using 18GB of elasticsearch memory.

If it's really the case that none of our queries are fielddata for anything, I suspect disabling it would at least save us memory, and probably improve performance.

riordan:
YESSSSSSSSSSSSSS

Binternet:
Just checked out the `disable_fielddata` branch and tried to run wof importer (production branch) and this popped up:

`error: [dbclient] [400] type=strict_dynamic_mapping_exception, reason=mapping set to strict, dynamic introduction of [empire] within [parent] is not allowed`

orangejulius:
Oh, good catch. I'll add that to the schema too!

orangejulius:
A final signoff here, this causes a huge reduction in memory usage for Elasticsearch and a nice performance boost. I haven't tested a full planet build yet, but here is a printout of fielddata memory usage on a 4-node cluster before this change. After this change, it will be 0 bytes.

```
id                     host           ip             node               total name.default
C5Byr9OxTDqnNcANncjqgQ 172.20.108.159 172.20.108.159 Rigellian Recorder 8.9gb        2.6gb
_XYaH099SYK5EHHtfb6W-g 172.20.112.152 172.20.112.152 Solarman             9gb        2.6gb
BO9TGTGmT8SfxZAaBdybWA 172.20.62.101  172.20.62.101  Andrew Gervais     8.9gb        2.6gb
nj-Z42eiT7W4KoxgJRiO6A 172.20.94.205  172.20.94.205  Miracle Man        8.8gb        5.3gb
```

So that's **36GB** of memory in the cluster that was previously wasted for no use at all

PR #302, opened 2018-06-08
Disable fielddata for name.* and phrase.* fields

orangejulius:
Fieldata is only used for aggregations, sorting (by the string value of a field), and scripts. Pelias does not use any of those.

Currently, a full planet Pelias build has about 40GB of fielddata loaded into memory, but it's never used. So this is a significant savings potential.

We'll want to test this on larger builds to get a sense of the memory savings and any other possible side effects.

Fixes https://github.com/pelias/schema/issues/100

### DIFF
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -137,7 +137,7 @@ var schema = {
         type: 'string',
         analyzer: 'peliasIndexOneEdgeGram',
         fielddata : {
-          loading: 'eager_global_ordinals'
+          format: "disabled"
         }
       }
     },
@@ -149,7 +149,7 @@ var schema = {
         type: 'string',
         analyzer: 'peliasPhrase',
         fielddata : {
-          loading: 'eager_global_ordinals'
+          format: "disabled"
         }
       }
     }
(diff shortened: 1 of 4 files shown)


orangejulius:
I've just rebased this branch and it is now up to date with master, and no longer built on top of #282

orangejulius:
A final signoff here, this causes a huge reduction in memory usage for Elasticsearch and a nice performance boost. I haven't run performance comparisons on a full planet build, but here is a printout of fielddata memory usage on a 4-node cluster before this change. After this change, it will is 0 bytes.

```
id                     host           ip             node               total name.default
C5Byr9OxTDqnNcANncjqgQ 172.20.108.159 172.20.108.159 Rigellian Recorder 8.9gb        2.6gb
_XYaH099SYK5EHHtfb6W-g 172.20.112.152 172.20.112.152 Solarman             9gb        2.6gb
BO9TGTGmT8SfxZAaBdybWA 172.20.62.101  172.20.62.101  Andrew Gervais     8.9gb        2.6gb
nj-Z42eiT7W4KoxgJRiO6A 172.20.94.205  172.20.94.205  Miracle Man        8.8gb        5.3gb
```

So that's **36GB** of memory in the cluster that was previously wasted for no use at all

PR #307, opened 2018-06-26
Split on whitespace for peliasStreet tokenizer

orangejulius:
This is just one small part of the changes originally in https://github.com/pelias/schema/pull/291

It should help especially with synonyms and paves the way for other improvements to the `peliasStreet` analyzer/tokenizer, while being small and easy to test.

@missinglink there were lots of tests with `match_phrase` added in #291, do you remember why? should we include them in this PR?

cc @adefarge

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -33,7 +33,7 @@ function generate(){
         },
         "peliasStreetTokenizer": {
           "type": "pattern",
-          "pattern": "[,/\\\\]+"
+          "pattern": "[\\s,/\\\\]+"
         }
       },
       "analyzer": {
--- a/test/fixtures/expected.json
+++ b/test/fixtures/expected.json
@@ -8,7 +8,7 @@
         },
         "peliasStreetTokenizer": {
           "type": "pattern",
-          "pattern": "[,/\\\\]+"
+          "pattern": "[\\s,/\\\\]+"
         }
       },
       "analyzer": {


missinglink:
I think it's important that we have some integration tests covering this change.

The `match_phrase` tests were in order to ensure that the behaviour would be correct *when using a phrase query*.

I am not sure if there are queries generated against `street` that don't use `phrase`, so we are probably best to add some tests for that too, to see what will happen.

I imagine that a query such as `street` will match half the index which is not ideal.

If that's the case then we will need to do a quick review of the queries we are generating in `pelias/api` to ensure that they are all using `phrase`.

missinglink:
`npm run integration` return `Failed Tests: There were 27 failures`

orangejulius:
wow, we don't run them: https://github.com/pelias/schema/blob/master/package.json#L16

ok, that needs to be fixed first :)

orangejulius:
Okay, in digging into this PR and https://github.com/pelias/schema/pull/310, I've realized it's not correct to separate them like this.

We originally split on whitespace so that we could handle synonyms that compose multiple words. like "beaverton hillsdale highway" and other Portland specific things. So merging this by itself would require us to re-do a bunch of testing without much benefit. I believe we should look at merging https://github.com/pelias/schema/pull/310 by testing it beyond just portland_metro, and we should close https://github.com/pelias/schema/pull/291

PR #310, opened 2018-08-02
Improved synonyms system

orangejulius:
This PR includes all the changes in #291 except for those which are Portland-specific, and not generally applicable.

It makes the following changes:
- Use whitespace to split tokens in the `peliasStreet` analyzer
- Move synonym definitions out of dynamically created filters and into synonym files, greatly reducing the size/complexity of the schema itself


missinglink:
I came across a relevant example today `17217 SW STARBUCK LN` in Portland. In this example `SW STARBUCK LN` is a valid street but `STARBUCK LN` is not a valid street anywhere in the world (as far as I could tell).

Maybe not an issue because `SW STARBUCK LN` is potentially a valid match for `STARBUCK LN`.

orangejulius:
It's been a while since I've thought about this issue. We really do need to decide whether or not to merge this soon. Can you expand on how your example is relevant? And is it an example of how this PR breaks things, how this PR helps, or something else?

orangejulius:
I took a look at what sort of impact this branch would have by running a North America build.

TLDR: the change in query results is not too bad, but the impact on performance is noticeably bad.

## Query result changes

Because we now tokenize on whitespace on `peliasStreets` analyzer used to index into and search against the `address_parts.street` field, we now can return matches that match all of the tokens in the query, as well as extras. Two examples of this, which are now failing tests are:

### [/v1/search?text=49 kay street](http://pelias.github.io/compare/#/v1/search%3Ftext=49%20kay%20street)
image

This query now returns results from `Kay Street` as well as `De Kay Street`. This is listed as an "unexpected" result in our acceptance tests, but it does score lower than the exact matches, so I believe it's not a huge deal

### [/v1/search?text=450 37th st, new york, ny 11232](http://pelias.github.io/compare/#/v1/search%3Ftext=450%2037th%20st,%20new%20york,%20ny%2011232)
image


This query now returns results for both `450 37th st` and `450 West 37th St`. Again, not a huge deal, and the more exact matches are shown first.

## Performance

However, as we feared, there a some obvious performance degradation from searching the index for common query terms like `street`. This only happens when there aren't enough uncommon other terms to lower the total number of matched documents. A particularly unperformant query from the acceptance tests is actually a venue query:

[/v1/search?text=Washington University in St Louis](http://pelias.github.io/compare/#/v1/search%3Ftext=Washington%20University%20in%20St%20Louis)

This query takes over 15 seconds to complete on a `c5.xlarge` (4 core) instance, with just north america data.

Looking at the Elasticsearch [profiler](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-profile.html) output, it's fairly clear where the problem lies: the total count of matched documents for the query is 42,912,211, or about 1/4 of all the documents in a North America build! 

Digging further, most parts of the query are very fast. For example, querying the name.phrase field for `washington university in (st street) louis` takes only 1.2 milliseconds. However, searching `address_parts.street` for _just_ `st` and `street` (both are searched due to synonyms) took 6767ms on a single shard:
```
                           "query_type": "BooleanQuery",
                           "lucene": "address_parts.street:st address_parts.street:street",
                           "time": "6767.610199ms",
```
Here's the full profile output: https://gist.github.com/orangejulius/ecd43607a059fa383c499a7b49c185e6

Curiously, a query that was fairly fast is [5 avenue street](http://pelias.github.io/compare/#/v1/search%3Ftext=5%20avenue%20street). I think this is because there are actually very few records that match both `avenue` and `street`

orangejulius:
Regarding performance, the "Washington University in St Louis" really is a uniquely bad case. 

Exploring other queries, even a query for "main st" is quite fast, matching 1.1M documents and taking around 180ms, or 300-400ms with profiling enabled ([profiler output](https://gist.github.com/orangejulius/d93614d978e1fe29a0a639ccc882c910)).

Furthermore, its possible we can improve query performance significantly without _too_ much work by doing a little work to detect queries with common tokens, and writing queries in a way that helps Elasticsearch match fewer documents.

For example, in the university query, we ended up with a `should` query like the following:
```
        {   
          "match": {
            "address_parts.street": {
              "analyzer": "peliasStreet",
              "boost": 5,
              "query": "Washington University in St"
            }   
          }   
        },  
```

If we had code to detect any extremely common tokens, we can help out Elasticsearch by rewriting that query clause as follows:
```
            {
              "bool": {
                "must": [
                  {
                    "match": {
                      "address_parts.street": {
                        "analyzer": "peliasStreet",
                        "boost": 5,
                        "query": "Washington University in"
                      }
                    }
                  },
                  {
                    "match": {
                      "address_parts.street": {
                        "analyzer": "peliasStreet",
                        "boost": 5,
                        "query": "Washington University in St"
                      }
                    }
                  }
                ]
              }
            },
```

This reduces the number of matched documents to only 2.1 million (down from 42 million!), and then the query takes only about 300ms.

This sort of approach could give us the following benefits:
 - allows us to detect expensive query terms and mitigate their cost, while still using them for matching
- in the case where the query is entirely common terms (for example, "avenue street"), we can keep the query as-is, instead of returning no results, as would happen if we used Elasticsearch stop words, or removed any expensive terms entirely

orangejulius:
I've run a full planet build with these changes, and did not notice any new acceptance test failures. The only thing left before being able to merge this is to do a little bit of performance comparison, with and without https://github.com/pelias/api/pull/1213 involved.

Joxit:
Hi there,

We are using this PR (customised) on a full planet build and it seems that we have no performance issues.
Here is the customised branch we use : [joxit/pelias-schema/build-2018-11](https://github.com/Joxit/pelias-schema/tree/build-2018-11).

I ran a full acceptance tests on both old prod (91.4%) and new one (89.51%), but errors are not relevant in our use-case.

BTW :+1: for ES5, works like a charm :smile:

orangejulius:
Nothing left to do :)

This PR looks good, and it's time for it to be merged. Looking forward to building on top of it in the future!

PR #319, opened 2018-09-26
refactor to more easily accommodate additional versions

missinglink:
Refactor to more easily accommodate additional versions

The idea with this PR is to version the existing elasticsearch 2.4 code so that we can add new versions for 5 and 6 in the future.
I have purposely not added any new versions at this time in order to keep the PR simple.

I would suggest that once a second config is added in the `./version` directory that we add some logic into `./schema.js` which selects the appropriate version to use at run-time.

The logic could be based off `pelias/config` (possibly the `esclient.apiVersion` property?) or via ENV var, I'll leave it up to the feature implementer to decide what they feel is best, maybe an ENV var is required to run 'development versions'.

I wasn't sure exactly where to "draw the line" between versions, so for now some things like `synonyms` are global, I'm happy to change this and version these things also if-and-when-required.

[edit] we also need to be cognizant of scripts bind-mounting or replacing the synonyms dictionary files and the effects of moving those files.

Similarly the integration tests are executed against whatever version has been loaded in the database, having these 'global' is useful in testing that they behave identically between versions 2 and 5, but might also need to be versioned when we look at a migration from 5 to 6.


missinglink:
```bash
# before
.
├── bin
├── integration
├── mappings
│   └── partial
├── scripts
├── synonyms
└── test
    ├── fixtures
    └── synonyms
```
```bash
# after
.
├── bin
├── integration
├── scripts
├── synonyms
├── test
│   ├── synonyms
│   └── version
│       └── 2.4
│           └── fixtures
└── version
    └── 2.4
        └── mappings
            └── partial
```

missinglink:
Yes, for now we might be able to get away with having a single version to rule them all :)

PR #321, opened 2018-09-27
remove geo_point properties deprecated in 2.3 or 2.4

missinglink:
remove geo_point properties deprecated in 2.3 or 2.4

- https://www.elastic.co/guide/en/elasticsearch/reference/2.4/geo-point.html#geo-point-params
- https://www.elastic.co/guide/en/elasticsearch/reference/2.4/release-notes-2.4.0.html

### DIFF
--- a/mappings/partial/centroid.js
+++ b/mappings/partial/centroid.js
@@ -1,14 +1,6 @@
 // @ref: http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/mapping-geo-point-type.html
 var schema = {
-  'type': 'geo_point',
-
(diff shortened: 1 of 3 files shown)


orangejulius:
Looks good to me, I didn't realize these were deprecated in 2.3/2.4.

Connects https://github.com/pelias/pelias/issues/461

orangejulius:
It's quite [difficult](https://github.com/elastic/elasticsearch/pull/20254) to enable the deprecation log in ES 2.x (it only logs at the debug level, and the debug level is very noisy), but I took a look and these deprecation warnings were the only ones there.

Therefore it looks like, as far as properties in the schema go, this is indeed the only change required for ES5 compatibility!

PR #326, opened 2018-10-08
Catch index already exists exception

gsoyka:
In our deployment, the schema container fails on the create_index job if the schema already exists in elasticsearch.  This patch resolves that issue and breaks the cycle of k8s constantly attempting to restart the job.

### DIFF
--- a/scripts/create_index.js
+++ b/scripts/create_index.js
@@ -18,8 +18,13 @@ var indexName = config.schema.indexName;
 
 client.indices.create( { index: indexName, body: schema }, function( err, res ){
   if( err ){
-    console.error( err.message || err, '\n' );
-    process.exit(1);
+    if( err.message.indexOf('index_already_exists_exception') > -1){
+      console.log( '[put mapping]: Index Already Exists', '\t', indexName, '\n' );
+      process.exit(0);
+    } else {
+      console.error( err.message || err, '\n' );
+      process.exit(1);
+    }
   }
   console.log( '[put mapping]', '\t', indexName, res, '\n' );
   process.exit( !!err );


orangejulius:
Hey @gsoyka,
Are you using the [pelias/kubernetes](https://github.com/pelias/kubernetes) helm charts? Our [schema build job](https://github.com/pelias/kubernetes/blob/master/build/templates/schema-create-job.tpl#L10-L21) has an option to drop the existing schema.

I'm a little hesitant to merge this because I believe failure to drop the schema or failure to create a schema are conditions that should be errors. For example, if changing the version of the schema that you are using, failure to create the schema means any new changes wouldn't be applied. Catching that error makes it easier to miss that.

I'm on the fence, what do you think?

gsoyka:
I am using the pelias/kubernetes helm charts.  Please forgive my ignorance around elasticsearch, but is it safe to drop and recreate an index, or will that result in data loss?  My main concern here is breaking the endless loop of kubernetes retrying the schema creation job if the index already exists, but I'm completely open to other ways of accomplishing that.

orangejulius:
Dropping the index _will_ delete all the data, so that's a valid concern. (among other things, `index` is the word Elasticsearch means in the same way Postgres or MySQL might use the word `database`, so it's analagous to dropping a database in an RDBMS) I admit myself to having spent a bit of time breaking that endless loop.

However, since the schema create job is used only to start a Pelias build, and it's [usually best](https://github.com/pelias/documentation/blob/master/pelias_from_scratch.md#aside-when-to-delete-the-data-already-in-elasticsearch) to start each build with a freshly-dropped index, usually losing that data has not been a concern.

Let me know if that matches with your workflows. I don't want to assume everyone does things the same way.

arne-cl:
I just ran into the same problem on Kubernetes (I'm not using HELM, but deployed the pelias-build jobs/pods manually).

Could the code for `./bin/create_index` be changed to only throw an error if the index exists AND it does
not match the schema that pelias needs? That way, the code can always run without "causing harm".

orangejulius:
Hey @arne-cl,
That's a really good idea.

I'm not sure if the schema JSON that we generate and send to Elasticsearch in this module is identical to what would come back if we ask Elasticsearch for the existing schema, but if you want to play around with it and see how much work it would be, that's definitely a feature we'd merge, and would solve this issue nicely.

PR #331, opened 2018-10-25
feat(mapping): use "index": "not_analyzed" for literal fields

### LINKED ISSUE #99: Explore using not_analyzed for source
The source field currently uses the "keyword" analyzer, which basically keeps the full string as a single token with no changes. According to [the keyword analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-keyword-analyzer.html) docs, it might make more sense to use the "not_analyzed" setting, briefly touched on [here](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/mapping-core-types.html) in the docs. It seems like it might do the same thing while somehow being faster.

orangejulius:
As long suspected in #99, turns out there _are_ differences between setting `"index": "not_analyzed"` for a field, and merely setting the analyzer to `keyword`.

They are detailed in the Elasticsearch 2.4 [String datatype](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/string.html#string-params) documentation, although it's a little bit confusing.

In Elasticsearch 5+, there are _two_ different types of string datatypes:

- [`text`](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/text.html)
- [`keyword`](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/keyword.html)

These documentation pages make the difference much more clear. In short, in Elasticsearch 2.4, setting `"index": "not_analyzed"` gives the following changes, all of which we'd like for these literal fields:

- Analysis is skipped all together, the raw value is added to the index
directly (this is pretty much equivalent to setting `analyzer: keyword`)
- [norms](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/norms.html) are disabled for the field, saving some disk space
- [doc_values](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/doc-values.html) are _enabled_.

The last one is most interesting. `doc_values` take up a little disk space but allow us to very efficiently perform aggregations. Pelias doesn't perform aggregations today. However, after we begin using a [single mapping type](#293), we will have to use aggregations in the [Pelias dashboard](https://github.com/pelias/dashboard) and any of our own analysis scripts to provide document counts for different sources or layers. The dashboard currently uses an API to get the count for each mapping type, which won't be supported going forward.

While minor, we needed a solution to this, and the only other one is field-data which is extremely expensive in terms of memory usage.

Test results
--------

In my testing, for the Portland metro Docker project, disk usage went from 451MB to 473MB, or about a 5% increase. The base for this comparison is #329, which should be merged first.

If we wanted to trim that down a bit, we could consider disabling `doc_values` for the `parent.*_id` fields. We don't have an immediate need for `doc_values` on those fields, although it might be interesting
for analysis. I quickly tested this out and it lowers disk usage further, another 4% to 430MB.

Summary
------

While not technically required for [Elasticsearch 5 support](pelias/pelias#461), this PR does bring us more in line with the best practices of ES5.

It also sets us up for [Elasticsearch 6](pelias/pelias#719) where the `string` datatype we use now is completely removed.

Fixes #99

### DIFF
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -3,14 +3,15 @@ const postalcode = require('./partial/postalcode');
 const hash = require('./partial/hash');
 const multiplier = require('./partial/multiplier');
 const literal = require('./partial/literal');
+const literal_with_doc_values = require('./partial/literal_with_doc_values');
 const config = require('pelias-config').generate();
 
 var schema = {
   properties: {
 
     // data partitioning
-    source: literal,
-    layer: literal,
+    source: literal_with_doc_values,
+    layer: literal_with_doc_values,
     alpha3: admin,
 
     // place name (ngram analysis)
--- a/mappings/partial/literal.json
+++ b/mappings/partial/literal.json
@@ -1,4 +1,5 @@
 {
   "type": "string",
-  "analyzer": "keyword"
+  "index": "not_analyzed",
+  "doc_values": false
 }
--- a/mappings/partial/literal_with_doc_values.json
+++ b/mappings/partial/literal_with_doc_values.json
@@ -1,0 +1,4 @@
+{
+  "type": "string",
+  "index": "not_analyzed"
+}
--- a/test/document.js
+++ b/test/document.js
@@ -117,7 +117,7 @@ module.exports.tests.parent_analysis = function(test, common) {
       t.equal(prop[field+'_a'].type, 'string');
       t.equal(prop[field+'_a'].analyzer, 'peliasAdmin');
       t.equal(prop[field+'_id'].type, 'string');
-      t.equal(prop[field+'_id'].analyzer, 'keyword');
+      t.equal(prop[field+'_id'].index, 'not_analyzed');
 
       t.end();
     });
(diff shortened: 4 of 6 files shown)


missinglink:
Let's discuss this more on a call, couple of notes:

- `not_analyzed` presumably only with with a `terms` query? We'd just need to check that we're not using any other wuery type for those fields such `match`
- we have some code to load dynamically load type_mappung information from the index (layers & sources), but it's currently disabled due to the aggregation taking several seconds on the first load, will `doc_values` improve this?

missinglink:
I think we might want to have `doc_values` enabled so we can do aggregations, seems relevant for all fields except `source_id`.

In particular aggregating on `source` and `layer` is useful, the rest would be purely for statistical purposes, such as `category` and the parent id fields.

orangejulius:
Yeah, we definitely need `doc_values` on `source` and `layer`.

I ran some Australia builds (14M documents) with variations on this change and the results are similar:

Configuration |  ES store size | change
-- |  -- | --
master |  | 6264mb | -
Change literals to `not_analyzed` |  6439mb | 2.79%
Disable `doc_values` on literals except sourece/layer |  6050mb | -3.41%

Given that we stand to save about 6% disk space by turning `doc_values` off where we aren't sure to need them, and that disk space translates directly into good things like faster snapshot save/load times, I'd say we go for it.

orangejulius:
I've updated this PR to include disabling `doc_values` on all fields except `source` and `layer`.

PR #337, opened 2018-12-04
Support Travis VM Infrastructure

orangejulius:
Building on the changes to Travis config in #336, this PR modifies our Elasticsearch setup to be compatible with VM based infrastructure. It's based heavily on the [Travis Elasticsearch](https://docs.travis-ci.com/user/database-setup/#elasticsearch) docs.

However, there currently appears to be an issue where for some reason, the Elasticsearch 5 builds think the Pelias index already exists. It almost looks as if the different builds are sharing data. However, that shouldn't be the case.

Since Travis has already started rolling out the switch to VM based builds, we have to fix this soon, and until then live with failing schema builds.

@missinglink any ideas?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
-sudo: false
 language: node_js
+# Once openjdk/oraclejdk dependencies are resolved, change to dist: xenial
+dist: trusty
 notifications:
   email: false
 node_js:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ env:
 jdk:
   - oraclejdk8
 install:
-  - mkdir /tmp/elasticsearch
   - ./scripts/setup_ci.sh
   - npm i
 script:
--- a/scripts/setup_ci.sh
+++ b/scripts/setup_ci.sh
@@ -4,29 +4,27 @@ set -e
 # download and install elasticsearch with ICU plugin
 # note: the download servers and plugin install binary changed between versions
 
+# default download and plugin locations
+ES_PLUGIN_BIN="/usr/share/elasticsearch/bin/elasticsearch-plugin"
+ES_DEB_URL="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.deb"
+
+# override if using old Elasticsearch versions
 if [[ "${ES_VERSION}" == "2.4"* ]]; then
-
-  # download from legacy host
-  wget -O - https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/${ES_VERSION}/elasticsearch-${ES_VERSION}.tar.gz \
-    | tar xz --directory=/tmp/elasticsearch --strip-components=1
-
-  # install ICU plugin
-  /tmp/elasticsearch/bin/plugin install analysis-icu
-
-  # start elasticsearch server
-  /tmp/elasticsearch/bin/elasticsearch --daemonize --path.data /tmp
-else
-
-  # download from new host
-  wget -O - https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${ES_VERSION}.tar.gz \
-    | tar xz --directory=/tmp/elasticsearch --strip-components=1
-
-  # install ICU plugin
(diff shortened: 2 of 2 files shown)


missinglink:
Hmm... Yeah that's super weird, I agree it looks like the VMs are sharing data between them.

I wonder if there is a new way of specifying the matrix config which enforces separation.

missinglink:
Agh.. so maybe this?

https://docs.travis-ci.com/user/caching/#caches-and-build-matrices

Any easy test to see if it's caching related us to set `cache: false` at the top of the Travis config, then work from there.

orangejulius:
I don't think it's actually caching. If you look at the [build output for a test i did](https://travis-ci.org/pelias/schema/jobs/463486794), there are no indices returned from the `cat` Elasticsearch API. And it only happens for ES5. Maybe there is weird behavior in ES5 we don't understand yet.

missinglink:
This SO issue might be related, looks like "index already exists" might be returned in cases which are not strictly true and that additional error info might not be showing in our debug output?

https://stackoverflow.com/questions/46400502/elasticsearch-create-index-index-already-exists-exception

orangejulius:
Okay, I figured it out. The duplicate index issue is a red-herring. It turns out Elasticsearch retries on index creation failure (as it does in all other requests, unless you set `maxRetries: 0` in the elasticsearch-js config.

Running a debug travis build, I did so, and saw the real error:

```
 PUT http://localhost:9200/pelias => Parse Error
      at Log.error (/home/travis/build/geocodeearth/pelias-schema/node_modules/elasticsearch/src/lib/log.js:226:56)
      at checkRespForFailure (/home/travis/build/geocodeearth/pelias-schema/node_modules/elasticsearch/src/lib/transport.js:262:18)
      at HttpConnector.<anonymous> (/home/travis/build/geocodeearth/pelias-schema/node_modules/elasticsearch/src/lib/connectors/http.js:163:7)
      at ClientRequest.wrapper (/home/travis/build/geocodeearth/pelias-schema/node_modules/lodash/lodash.js:4935:19)
      at ClientRequest.emit (events.js:182:13)
      at Socket.socketOnData (_http_client.js:447:9)
      at Socket.emit (events.js:182:13)
      at addChunk (_stream_readable.js:283:12)
      at readableAddChunk (_stream_readable.js:264:11)
      at Socket.Readable.push (_stream_readable.js:219:10)

{ Error: Parse Error
    at Socket.socketOnData (_http_client.js:441:20)
    at Socket.emit (events.js:182:13)
    at addChunk (_stream_readable.js:283:12)
    at readableAddChunk (_stream_readable.js:264:11)
    at Socket.Readable.push (_stream_readable.js:219:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:94:17) bytesParsed: 8308, code: 'HPE_HEADER_OVERFLOW' } '\n'
```

missinglink:
Face
Palm

orangejulius:
Okay, I think I understand now. I don't believe this particular error is related to Travis VMs at all. I was able to reproduce the error locally on all versions of Node.js but the latest.

The 8192 byte header limit was added in the [recent security fix Node.js releases](https://nodejs.org/en/blog/vulnerability/november-2018-security-releases/), and it appears that Node.js is now adding some followup PRs to do things like [allow configuring the limit on the command line](https://github.com/nodejs/node/pull/24811), which would help us here.

There is also the unfortunate fact that while Elasticsearch 5 has an http option to limit the total size of headers generated, it doesn't seem to work in my testing. There _are_ actually new parameters introduced to control deprecation warning headers, but they were [only introduced in Elasticsearch 6.3](https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-6.3.0.html) and don't appear to be backported to 5.

orangejulius:
I've updated this PR to change the Node.js versions to specify 8.13.0 and 10.13.0, the latest version in each release line without the header size limit, and everything now works!

I've opened https://github.com/elastic/elasticsearch/issues/36243 to see if Elasticsearch would consider backporting the changes that can bring us to the latest Node.js version safely. If they do, we can use that version and at least set the defaults in our Docker images.

Issue #311, opened 2018-08-07
lint javascript


missinglink:
for some reason code in this repo doesn't seem to conform to our usual jslint config.

just noticed this code:

```
if( isForced() ) drop();
else prompt( drop, fail );
```

PR #344, opened 2019-02-08
chore(test): Add precommit-hook

orangejulius:
This was never set up, and thus we have not been linting code in this
repository.

Fixes https://github.com/pelias/schema/issues/311


orangejulius:
This PR has been sitting in a branch for months. I think there was a reason why we couldn't merge it before, but it looks ok now.

orangejulius:
Ah, now I know the reason. There are a bunch of liter errors which will take some work to fix. I added a WIP commit that takes care of many of them. We can finish it off later.

Issue #285, opened 2018-05-07
Use Elasticsearch fields feature


orangejulius:
Currently, we take the "name" field (and all its language variants) and run it through Elasticsearch twice. Once with the "name" analyzers, and once in the "phrase.*" field with the phrase analyzers. This is done so that we can have special analysis to handle all the different use cases of autocomplete and regular search.

It works well, but is a bit of a pain to manage.

It turns out Elasticsearch has functionality to support exactly this functionality: [fields](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/multi-fields.html).

This allows one field to be indexed in multiple ways, without the confusion of multiple fields that do not have any inherent relationship in Elasticsearch. While there would have to be some cosmetic changes to all our queries, it looks like the change is not a big deal overall.

orangejulius:
Here's a full writeup of the changes that are likely required:

# Elasticsearch Multi-Field Refactor

A great introduction to the Pelias API, Elasticsearch, and general query logic is to take on a
refactoring task we've [wanted to do for a while](https://github.com/pelias/schema/issues/285).

## Background

To facilitate both search and autocomplete, we break down all our text data into two separate full
text indices. They are called the `name` and `phrase` fields and are roughly defined as follows:

`name`: input text is broken into many individual tokens of progressive characters of the input. for
example `main st` would be converted into the array `[m, ma, mai, main, s, st]`. This is necessary
for autocomplete.

`phrase`: input text is broken into an ordered list of words (roughly, since we split on more than
just whitespace). `main st` becomes `[0:main, 1:st]`. This is used both for autocomplete and search

We do this by actually having [two separate fields](https://github.com/pelias/schema/blob/master/mappings/document.js#L15-L19) in our
Elasticsearch schema. From Elasticsearch's perspective, there is no relation between these two
fields. Only our application code in the [Pelias API](https://github.com/pelias/api) understands the
relationship.

However, although we didn't know it at the time, Elasticsearch has a feature to allow one input text
to be analyzed [multiple ways](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/multi-fields.html).

Using this feature would be a pure refactoring from a functionality standpoint (all queries should
return exactly the same results as before) but should lead to increased readability in our code, and
probably some minor disk usage savings and possibly performance increases in Elasticsearch.

## Pelias Components Involved

### [Pelias schema library](https://github.com/pelias/schema)

This is where our code for managing Elasticsearch's schema lives. It's a set of
[...]

PR #345, opened 2019-02-08
Compute an ngram field for all admin data

missinglink:
This PR adds one new field per `parent.*` admin entry called, for example `parent.locality_ngram`.

It takes advantage of the `copy_to` operation in order to copy the admin name and admin abbreviation inputs for each record and adds them to the new ngram field.

The ngram field is tokenized using the existing `peliasIndexOneEdgeGram` analyzer, so as to produce prefix-ngrams which can be used for autocomplete.

The motivation here is to be able to, quite simply and efficiently, improve autocomplete queries which contain admin areas.
We currently only autocomplete on the `name.default` field and require full completion of the admin inputs before any effect is noticed in the results.

I suspect the changes required for the queries (such as [this](https://github.com/pelias/api/blob/master/test/unit/fixture/autocomplete_linguistic_with_admin.js)) will be minimal.
Due to the way `addressit` works (splitting 'address parts' and 'admin parts') this should play well with the existing parsing logic.


missinglink:
@orangejulius there will certainly be additional disk overhead for these extra fields, but due to the repetitive nature of the data in these fields, I suspect that it won't be that noticeable once converted to an inverted-index.

I disabled `field-data` and `doc_values` so that should also reduce the index size.

One nice unexpected benefit of using `copy_to` was that we don't need to add these fields to `_source.excludes` because they contain no `_source` data.

I'm not 100% sure about the performance impact, especially common terms such as `new` or `west`, so we'll need to do a global-build test to check that it doesn't kill performance.

missinglink:
I did a build of `portland-metro` and compared disk usage to `master`:

```
➜  portland-metro du -sh elasticsearch
423M	elasticsearch

```

orangejulius:
Ouch, yeah that disk space usage would be a problem. Try using the [fields feature](https://github.com/pelias/schema/issues/285) which is even more well suited to storing multiple analysis variations for a single field than `copy_to`

missinglink:
@orangejulius I rewrote this PR using the `fields` feature:

```
➜  portland-metro du -sh elasticsearch
459M	elasticsearch
```

I'm not sure why this uses much less HDD space but it's great, I will open a separate PR.

missinglink:
Closing issue, this PR has been superceded by https://github.com/pelias/schema/pull/345

orangejulius:
That's awesome that it saves a bunch of space and makes this feature possible. Perhaps we should use it to replace `phrase` as well.

PR #350, opened 2019-03-04
add name synonyms for common languages

missinglink:
This PR adds `name.default` synonyms for some common languages.

We already have synonyms for `admin`, `street` etc but currently none for `name`.

This PR is aimed at fixing the reported bug where the venue `Mount Tabor Park` was not retrievable using abbreviated forms of the word `Mount`.

The synonyms were sourced from `libpostal`


missinglink:
ping @Joxit, what do you think?

orangejulius:
I just tested this out with a full planet build we kicked off. It looks to fix the mt/mount issue with no visible side effects.

In particular, it fixes the "mt zion" search query in https://github.com/pelias/pelias/issues/256. I've added some tests in https://github.com/pelias/acceptance-tests/pull/491 for this

orangejulius:
I think this is good to merge :)

PR #355, opened 2019-03-22
update Germanic street suffix synonyms

missinglink:
A user reported that `Aufderhöher Str` is not synonymous with `Aufderhöher Straße` and `Aufderhöher Strasse`.
This PR adds street suffix synonyms for common Germanic street names.

It doesn't handle contracted and expanded version of compound words as per `Aufderhöherstr.` or `Aufderhöherstraße`.


orangejulius:
Looks good to me. Other than `platz/pl/place`, I don't see any obvious cases of overlap or potential for undesired behavior. I honestly think this is probably good to merge and we'll see the effects in the next build.

PR #359, opened 2019-04-29
remove full_token_address_suffix_expansion

missinglink:
We currently have a synonyms file called `full_token_address_suffix_expansion` which I believe we can remove in favour of using the combination of two existing synonyms files (`street_suffix` & `directionals`).

I can't remember the history here but the `full_token_address_suffix_expansion` synonyms are presumably intended for use when the `full_token` is available (ie. not for partially complete tokens)

Despite this, the `full_token_address_suffix_expansion` functionality seems to avoid 'unsafe' expansions which would only be relevant for partially complete tokens.

The synonyms also follow the legacy synonyms format where tokens were 'expanded' from their contracted form to their expanded form (ie. a 'replacement'), the more recent approach we've adopted is to use the `,` convention in mappings rather than the `=>` convention which allows for either the contracted or expanded form to match.

This feels good to me but I'm a little uneasy about it simply because I don't remember why it was the way it was, I think it's just because the code is old and it's better off being removed and replaced.


missinglink:
@orangejulius this should fix the issue you mentioned recently where the following two queries are not synonymous:
- https://pelias.github.io/compare/#/v1/autocomplete%3Fboundary.gid=whosonfirst:region:85688637&text=mcalby%20ct%20murietta%20CA
- https://pelias.github.io/compare/#/v1/autocomplete%3Fboundary.gid=whosonfirst:region:85688637&text=mcalby%20court%20murietta%20CA

missinglink:
This PR is ready for testing and merging. Originally I coupled this with another PR but I now think it's best to merge on its own.
GTG

PR #364, opened 2019-05-20
Compute an ngram field for all street names

missinglink:
This PR adds one new field called `address_parts.street.ngram`

It takes advantage of the `fields` mapping to generate ngrams for street names.

The ngram field is tokenized using a new tokenizers called `peliasIndexStreetOneEdgeGram` which is mostly the same as the `peliasIndexOneEdgeGram` analyzer except using different synonyms files, so as to produce prefix-ngrams which can be used for autocomplete.

The motivation here is to be able to, quite simply and efficiently, improve autocomplete queries which contain street names.
We currently do autocomplete on the `name.default` field, which mixes names and addresses, using this field with a `multi_match` will have benefits over the single-field approach:

- Allow per-field analysis including synonym substitutions
- Allow for the fields to be included or excluded at query-time using a `multi_match` query

I suspect the changes required for the queries (such as this) will be minimal.

It'd likely that the index size will be increased after merging this PR because the street names will be indexed using an edge ngram filter twice, once for `name.default` and once for `street.ngram`.

The plan is, following this PR to remove the street names from the `name.default` field, once this has been done the index will return to the previous size (or a very similar on-disk size).

The analysis that I've configured in this PR is likely not perfect, but it mirrors what we already have, so the integration will be easier.
Once we've merge this PR and switched the queries we will be much freer to improve individual fields and analysis.

related: https://github.com/pelias/schema/pull/347
related: https://github.com/pelias/schema/pull/359

### DIFF
--- a/integration/analyzer_peliasIndexStreetOneEdgeGram.js
+++ b/integration/analyzer_peliasIndexStreetOneEdgeGram.js
@@ -1,0 +1,192 @@
+// validate analyzer is behaving as expected
+
+var tape = require('tape'),
+  elastictest = require('elastictest'),
+  schema = require('../schema'),
+  punctuation = require('../punctuation');
+
+module.exports.tests = {};
+
+module.exports.tests.analyze = function (test, common) {
+  test('analyze', function (t) {
+
+    var suite = new elastictest.Suite(common.clientOpts, { schema: schema });
+    var assertAnalysis = analyze.bind(null, suite, t, 'peliasIndexStreetOneEdgeGram');
+    suite.action(function (done) { setTimeout(done, 500); }); // wait for es to bring some shards up
+
+    assertAnalysis('lowercase', 'F', ['f']);
+    assertAnalysis('asciifolding', 'á', ['a']);
+    assertAnalysis('asciifolding', 'ß', ['s', 'ss']);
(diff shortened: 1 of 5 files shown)


missinglink:
This is the plan to roll this out without breaking backwards compatibility:

1. Update schema to add `street.ngram` field
2. Update queries to use `multi_match`
3. Queries should now work with street data *either* in `name.default` or `street.ngram` (backwards and forwards compatible)
4. Check the codebase to ensure that Documents with no name are considered valid
5. Remove [osm](https://github.com/pelias/openstreetmap/blob/master/stream/address_extractor.js#L66) and [oa](https://github.com/pelias/openaddresses/blob/master/lib/streams/documentStream.js#L26) code which concats the housenumber and street name together.
6. Remove `name.default` from `multi_match` where applicable (remove backwards compatibility)

missinglink:
This is a really cool feature but I didn't realise how much work would be involved in order to change the way `name.default` works.
In particular, there would need to be changed to how search works and also how labels are generated.

Let's leave this open for discussion, I'd still very much like to merge this one day because it's a big step forward for modernizing our schema design based on learnings over the last few years.

missinglink:
The increase in disk space for the whole planet was >20GB

PR #367, opened 2019-06-07
unique token filter bugfix

missinglink:
I discovered a bug today which can cause `match_phrase` queries to fail in certain situations.

I was investigating the query `SE 23rd Ave & SE Madison St` and discovered that the `unique` token filter was stripping the first `SE` token, and since the second `SE` token position was too far away according to the `slop` value, it failed to match the input `23rd Ave & SE Madison St`.

A simplified example is something like this (with token positions shown):
```
input: 'A AB'

['0:A', '0:AB'] // before
['0:A', '1:A', '1:AB'] // after
```

The first 'A' ngram would be removed and also the other token positions would be incorrect as a result!

I'm not sure about the full extent of how many queries this affects but I suspect it causes jitter issues.
Only `phrase` type queries are affected and it's only really relevant for `ngram` analysis.

This PR switches all analyzers to use a new `unique_only_same_position` filter which uses the `only_on_same_position` option.

~~I kind of regret it (because the PR is harder to read), but I also changed the way the tests are run to remove duplication in the test code, so it added some noise to this PR.~~

### DIFF
--- a/integration/analyzer_peliasIndexOneEdgeGram.js
+++ b/integration/analyzer_peliasIndexOneEdgeGram.js
@@ -23,17 +23,20 @@ module.exports.tests.analyze = function(test, common){
     assertAnalysis( 'trim', ' f ', ['f'] );
     assertAnalysis( 'ampersand', 'a and b', ['a','&','b'] );
     assertAnalysis( 'ampersand', 'a & b', ['a','&','b'] );
-    assertAnalysis( 'ampersand', 'a and & and b', ['a','&','b'] );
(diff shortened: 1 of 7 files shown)


orangejulius:
I couldn't pass up the chance to use some git-fu, so the test refactoring has now been extracted into https://github.com/pelias/schema/pull/368 and this PR now includes only the changes required to use the `only_on_same_position` flag for the unique token filter. 🎉

missinglink:
Any blockers on merging this @orangejulius ?

orangejulius:
Nope, let's merge :)

PR #368, opened 2019-06-07
chore(test): Use more common tools in tests

orangejulius:
This refactoring was extracted from the bugfix in https://github.com/pelias/schema/pull/367


missinglink:
Nice :+1:

orangejulius:
Yup, a little git magic and I was able to extract out just the test changes.

PR #370, opened 2019-06-10
Simplify query-time analysis

missinglink:
We have two 'query-time' analyzers right now:
- peliasQueryPartialToken
- peliasQueryFullToken

The difference between these analyzers is due to how the synonyms substitutions are handled.

Some time ago we made a change to how synonyms are applied in schema, so instead of 'expanding' or 'contracting' like we used to do, we now allow pure synonym substitution where either/or version of the token will match.

So this means that query-time synonym generation is no longer required as all versions are available in the index.

I started doing the work to remove the synonyms filter and realised quickly that `peliasQueryPartialToken` and `peliasQueryFullToken` would now do exactly the same thing, albeit under a different name.

I would like to remove both of those analyzers as they are just confusing and don't serve any functional purpose, plus in their current form they will have a negative performance impact.

Removing or renaming the analyzers would be a breaking change for users who are using the old analyzer names in their [defaults configs](https://github.com/pelias/api/blob/master/query/search_defaults.js#L20).

So this PR adds a new analyzer (ironic I know!), which we can migrate the API code to use, and then later, after a courtesy window we can remove the old analyzers.

note: I wouldn't merge this until after the [ampersand](https://github.com/pelias/schema/pull/369) PR is merged, because that is the last synonyms file which does expansion/contraction

### DIFF
--- a/integration/analyzer_peliasQuery.js
+++ b/integration/analyzer_peliasQuery.js
@@ -1,0 +1,121 @@
+// validate analyzer is behaving as expected
+
+var tape = require('tape'),
+    elastictest = require('elastictest'),
+    schema = require('../schema'),
+    punctuation = require('../punctuation');
+
+module.exports.tests = {};
+
+module.exports.tests.analyze = function(test, common){
+  test( 'analyze', function(t){
+
+    var suite = new elastictest.Suite( common.clientOpts, { schema: schema } );
+    var assertAnalysis = common.analyze.bind( null, suite, t, 'peliasQuery' );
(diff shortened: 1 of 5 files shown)


missinglink:
It might also be nice to set the [search-analyzer](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-analyzer.html) for each field, this would remove the need to specify one at query-time.

missinglink:
Now that https://github.com/pelias/schema/pull/369 has been merged we can also remove the ampersand substitution at query time.

PR #372, opened 2019-07-05
Update schema for Elasticsearch 5 and 6 

orangejulius:
This change makes our Elasticsearch schema compatible with Elasticsearch 5 and 6. It shouldn't have any effect on performance or operation, but it will completely drop compatibility for Elasticsearch 2.

The primary change is that Elasticsearch 5 introduces two types of text fields: `text` and `keyword`, whereas Elasticsearch 2 only had 1: `string`.

Roughly, a `text` field is for true full text search and a `keyword` field is for simple values that are primarily used for filtering or aggregation (for example, our `source` and `layer` fields). The `string` datatype previously filled both of those roles depending on how it was configured.

Fortunately, we had already roughly created a concept similar to the `keyword` datatype in our schema, but called it `literal`. This has been renamed to `keyword` to cut down on the number of terms needed

One nice effect of this change is that it removes all deprecation warnings printed by Elasticsearch 5. Notably, as discovered in https://github.com/pelias/schema/pull/337#issuecomment-444313941, these warnings were quite noisy and required special handling to work around Node.js header size restrictions. This special handling can now been removed.

Fixes pelias/whosonfirst#457
Connects pelias/pelias#719
Connects pelias/pelias#461

### DIFF
--- a/bin/integration
+++ b/bin/integration
@@ -4,4 +4,4 @@
 # see https://github.com/pelias/pelias/issues/744
 set -o pipefail
 
-node --max-http-header-size=16384 integration/run.js | npx tap-spec
+node integration/run.js | npx tap-spec
--- a/integration/dynamic_templates.js
+++ b/integration/dynamic_templates.js
@@ -129,19 +129,12 @@ function addendumAssertion( type, namespace, value, common ){
         var addendumProperties = properties.addendum.properties;
 
         t.true([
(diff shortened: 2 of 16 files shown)


orangejulius:
I've now tested this fairly extensively across several small builds, and can't identify any regressions. The schema is expected to still be equivalent for ES5+, so this is now ready for full planet testing.

PR #133, opened 2016-06-07
Add complicated house number filter

trescube:
DO NOT MERGE, this work has been merged in to the `elasticsearch2` branch

---

This remarkably convoluted series of Lucene token filters is bundled into an analyzer called `goldbergianPeliasHousenumber` and allows us to remove house numbers from prefix-ngramming logic.  

Here's an example, `14a w main st`:

```
surround_single_characters_with_!s  -> [14a, !w!, main, st]
house_number_word_delimiter         -> [14, 14a, a, !w!, main, st]
remove_single_characters            -> [14, 14a, !w!, main, st]
surround_house_numbers_with_!s      -> [!14!, !14a!, !w!, main, st]
peliasOneEdgeGramFilter             -> [!, !1, !14, !14!, !, !1, !14, !14a, !14a!, !, !w, !w!, m, ma, mai, main, s, st]
eliminate_tokens_starting_with_!    -> ["", "", "", !14!, "", "", "", "", !14a!, "", "", w, m, ma, mai, main, s, st]
remove_encapsulating_!s             -> ["", "", "", 14, "", "", "", "", 14a, "", "", w, m, ma, mai, main, s, st]
notnull                             -> [14, 14a, w, m, ma, mai, main, s, st]
unique                              -> [14, 14a, w, m, ma, mai, main, s, st]
```

PR #379, opened 2019-09-25
Remove complicated housenumber filter

orangejulius:
This removes the complex filtering logic known as the "complicated housenumber filter", or sometimes even the "goldbergian housenumber filter", originally described in https://github.com/pelias/schema/pull/133 and merged as part of our ES2 upgrade (https://github.com/pelias/schema/pull/138).
    
The housenumber filter was made quite a long time ago, with the intention of removing partial numeric tokens.

It's a lot of code, rather hard to understand, and may actually cause more problems than it solves.

Since this code was written, we've improved autocomplete considerably, and having some tokens missing could be resulting in autocomplete not doing its job as well as it could.

In particular, fully numeric postal codes cannot be found with autocomplete until the postal code is entered completely (see https://github.com/pelias/pelias/issues/676).

More examples where this might have an impact:

* [30 W 2](https://pelias.github.io/compare/#/v1/search%3Ftext=30%20w%202) can only ever return addresses like 30 W 2nd st, never 30 W 26th st.
* [Reichenberger Str 6](https://pelias.github.io/compare/#/v1/search%3Ftext=reichenberger%20str%206) can return addresses with housenumber 6, but never 60, 61, 61b, etc

However, there could also be downsides: this will probably increase the hit count for numeric inputs dramatically, so it could cause more slow queries.

This isn't super high priority,  but this code has frustrated us for a long time, and was actually quite easy to delete, so we should explore whether or not it is an improvement.

### DIFF
--- a/integration/analyzer_peliasIndexOneEdgeGram.js
+++ b/integration/analyzer_peliasIndexOneEdgeGram.js
@@ -159,7 +159,7 @@ module.exports.tests.functional = function(test, common){
     ]);
 
     assertAnalysis('address', '101 mapzen place', [
-      '0:101', '1:m', '1:ma', '1:map', '1:mapz', '1:mapze', '1:mapzen', '2:p', '2:pl', '2:pla', '2:plac', '2:place'
+      '0:1', '0:10', '0:101', '1:m', '1:ma', '1:map', '1:mapz', '1:mapze', '1:mapzen', '2:p', '2:pl', '2:pla', '2:plac', '2:place'
     ]);
 
     suite.run( t.end );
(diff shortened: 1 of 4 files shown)


missinglink:
I'm nervous about releasing this, are there any acceptance tests regressions?

orangejulius:
I'm a bit nervous as well, who knows what we'll see. I'll launch a full planet build to test it out.

orangejulius:
I tested out a full planet build with this change today. There were very minor, if any differences, even comparing the full output of the acceptance tests in autocomplete mode (over 9000 queries).

Evaluating some individual queries that we _thought_ might have changed (such as the two listed above and a query for a partial postal code) shows that in almost all cases, the new pelias parser is generating queries against non-ngram fields (such as `address_parts.zip` or `phrase.default`), so this change has no effect.

A little bit of performance testing would still be warranted, but it really seems like this PR will have no negative side effects, while dramatically simplifying the schema code, fixing some odd issues with phrase ordering, and hopefully set us up for more predictable work to improve autocomplete at the query level.

orangejulius:
One more thing we should test here: housenumbers with unit numbers or other non-numeric characters in them. We actually have very few acceptance tests with these cases, but they are a common source of bugs.

orangejulius:
I tested this extensively yesterday including creating a new test suite for housenumbers with non-numeric characters (https://github.com/pelias/acceptance-tests/pull/509), and could not find _any_ differences in how Pelias behaves with this change.

Mostly, it seems that the way the autocomplete queries generated by the Pelias parser behave, any queries that would be impacted by new tokens in the index for housenumbers are not able to use them. We should follow up with some API changes to improve the queries we generate, but for now it means this PR is at the very least not hurting anything.

It's currently undergoing performance testing and unless the results are very concerning, it's likely we'll merge this soon!

orangejulius:
After a day of testing this PR is showing potentially improved performance against real workloads. It's not major, but at the very least there don't seem to be any noticeable performance regressions.

Looking at the number of Elasticsearch hits across about a million queries, there is also no noticeable change. So our fears of a massive increase in hits for short numeric tokens don't appear to have been realized.

missinglink:
Were there any significant changes to the disk requirements from the additional numeric prefix tokens?

orangejulius:
Nope, no changes in disk size at all:
Screenshot_2019-11-07_10-07-04

PR #385, opened 2019-10-30
elasticsearch 6

missinglink:
ES6 compatibility (not compatible with ES5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,9 @@ matrix:
   fast_finish: true
 env:
   matrix:
-    - ES_VERSION=5.6.12 JDK_VERSION=oraclejdk8
     - ES_VERSION=5.6.12 JDK_VERSION=oraclejdk11
-    - ES_VERSION=6.8.4 JDK_VERSION=oraclejdk8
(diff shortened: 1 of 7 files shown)


missinglink:
I don't think we can merge this until we officially move to ES6, there are some settings such as the removal of `_all` which would cause regressions on ES5.

orangejulius:
This looks good. Like you said lets hold off on merging this for a little while. At least for the core changes that will drop ES5 compatibility in the schema (the refactoring and CI changes could be extracted from this PR and merged at any time).

And we definitely need to remember to make sure this is a breaking change. I'd suggest putting it in the commit message now so we don't forget. :+1:

missinglink:
I'm closing this in favour of https://github.com/pelias/schema/pull/394

As this is PR not backwards compatible with ES5, there is no real value in moving to ES6 in a separate step from supporting ES7.

What we can do instead is move directly to a version of schema which both ES6+ES7 compatible and save a bunch of time and energy in the process.

The CI is set up for the PR above which tests against both versions 6 & 7 (and is passing).

PR #386, opened 2019-10-30
enforce strict checking of token positions

missinglink:
okay, so there are a bunch of changes in here but it should be a no-op refactor ;)

this PR enforces strict checking of token positions for the `integration` test suite.

the motivation for this is:
- the order in which tokens are generated by elasticsearch changed in elasticsearch@6 making all the tests break
- not specifying token positions in the test meant they were not checked, meaning nasty bugs could slip through unnoticed.

you can provide the 'expected' tokens in two forms:

```
// no explicit token positions
[ 'A', 'ABC' ]

or...

// with explicit token positions
[ '0:A', '1:ABC' ]
```

prior to this PR if you used the 'no explicit token positions' form then we simply checked that all the tokens were present and there were no extra ones being generated.

what it's doing now is assigning a token position equal to the array offset, so 'A' is being seen as '0:A' internally now, and all token positions are being checked.

the use of the other form goes unchanged but the requirement to add `, true` to indicate this style is now removed and no longer required.

I think this is a better approach going-forward as it prevents bugs, it does add some verbosity to testing ngrams and synonyms analysis but that's totally worth it in order to ensure it's working as expected :)

Oh and you can also specify the expected tokens out-of-order or in any order as long as the positions are correct.

### DIFF
--- a/integration/analyzer_peliasIndexOneEdgeGram.js
+++ b/integration/analyzer_peliasIndexOneEdgeGram.js
@@ -16,62 +16,72 @@ module.exports.tests.analyze = function(test, common){
 
     assertAnalysis( 'lowercase', 'F', ['f']);
     assertAnalysis( 'asciifolding', 'á', ['a']);
-    assertAnalysis( 'asciifolding', 'ß', ['s','ss']);
-    assertAnalysis( 'asciifolding', 'æ', ['a','ae']);
+    assertAnalysis( 'asciifolding', 'ß', ['0:s','0:ss']);
+    assertAnalysis( 'asciifolding', 'æ', ['0:a','0:ae']);
     assertAnalysis( 'asciifolding', 'ł', ['l']);
     assertAnalysis( 'asciifolding', 'ɰ', ['m']);
     assertAnalysis( 'trim', ' f ', ['f'] );
-    assertAnalysis( 'ampersand', 'a and b', [
+    assertAnalysis('ampersand', 'a and b', [
       '0:a',
       '1:a', '1:an', '1:and', '1:&',
       '2:b'
-    ], true );
-    assertAnalysis( 'ampersand', 'a & b', [
+    ]);
+    assertAnalysis('ampersand', 'a & b', [
       '0:a',
       '1:&', '1:a', '1:an', '1:and', '1:u', '1:un', '1:und',
       '2:b'
-    ], true );
-    assertAnalysis( 'ampersand', 'a and & and b', [
+    ]);
+    assertAnalysis('ampersand', 'a and & and b', [
       '0:a',
       '1:a', '1:an', '1:and', '1:&',
       '2:&', '2:a', '2:an', '2:and', '2:u', '2:un', '2:und',
       '3:a', '3:an', '3:and', '3:&',
       '4:b'
-    ], true );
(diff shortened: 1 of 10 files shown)


integration/analyzer_peliasIndexOneEdgeGram.js
```diff
@@ -175,15 +189,23 @@ module.exports.tests.address = function(test, common){
...
+      '3:s', '3:st', '3:str', '3:stre', '3:stree', '3:street'
     ]);
 
     assertAnalysis( 'address', '4B 921 83 st', [
-      '4b', '921', '83', 's', 'st', 'str', 'stre', 'stree', 'street'
+      '0:4b',
+      '2:921', // @todo: this token position is incorrect
+      '3:83',
+      '4:s', '4:st', '4:str', '4:stre', '4:stree', '4:street'
     ]);
```
missinglink: in the process of doing this PR I found one peculiarity, which is a good example of why this is important.

integration/run.js
```diff
@@ -20,20 +21,45 @@ const common = {
...
         index: suite.props.index,
-        analyzer: analyzer,
-        text: text
-      }, ( err, res ) => {
-        if( err ){ console.error( err ); }
-        t.deepEqual( common.simpleTokens( res.tokens, includePosition ), expected, comment );
+        body: {
+          analyzer: analyzer,
+          text: text.toString()
+        }
```
missinglink: this change was also interesting and is required in ES6.

I believe prior to this change the request was being made via query-string and now it's being made in the request body.

The interesting bit is when I change this some of the tests which handle commas started failing, so I think there is either a bug in the elasticsearch-js handling of this or something else weird was going on.

Regardless, this is the better form to use.

integration/analyzer_peliasIndexOneEdgeGram.js
orangejulius: woah very strange. any idea how or why this is happening?

integration/run.js
missinglink: Check this out, the second token position changes depending on the request method:

```js
GET http://localhost:9200/pelias/_analyze?analyzer=peliasIndexOneEdgeGram&text=a%2Cb

{
    "tokens": [
        {
            "token": "a",
            "start_offset": 0,
            "end_offset": 1,
            "type": "word",
            "position": 0
        },
        {
            "token": "b",
            "start_offset": 2,
            "end_offset": 3,
            "type": "word",
            "position": 101
        }
    ]
}
```

```js
GET http://localhost:9200/pelias/_analyze
{
	"analyzer": "peliasIndexOneEdgeGram",
	"text": "a,b"
}

{
    "tokens": [
        {
            "token": "a",
            "start_offset": 0,
            "end_offset": 1,
            "type": "word",
            "position": 0
        },
        {
            "token": "b",
            "start_offset": 2,
            "end_offset": 3,
            "type": "word",
            "position": 1
        }
    ]
}
```

missinglink: They removed the query-string version in ES6, so it's probably a known bug.

orangejulius: Pretty wild. So it has different behavior in ES6, but I assume the two queries would do the same thing in ES5? Or is there something else going on here?

Issue #381, opened 2019-09-27
Correct our use of synonyms for ES6


orangejulius:
While testing ES6 support, I ran into the following error while running an OSM import:

```
type=illegal_argument_exception, reason=startOffset must be non-negative, and endOffset must be >= startOffset, and offsets must not go backwards startOffset=9,endOffset=20,lastStartOffset=14 for field 'name.default'
```

After some digging it appears this [error](https://github.com/elastic/elasticsearch/issues/32694) is related to token position offsets created by the [Synonym token filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html).

There is a very interesting [Elastic blog post from 2017](https://www.elastic.co/blog/multitoken-synonyms-and-graph-queries-in-elasticsearch) discussing the solution: the new [Synonym graph token filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-graph-tokenfilter.html) and how to use it to improve how synonyms expansion works. We'll need to figure out what the right solution is here for ES6 support

Overall, the biggest takeaway appears to be this:

> To make multi-token synonyms work correctly you must apply your synonyms at query time, not index-time, since a Lucene index cannot store a token graph.

Connects https://github.com/pelias/pelias/issues/719

missinglink:
Hmm.. very interesting, I don't think we could get away with doing our synonyms at query-time because of autocomplete.
eg. if the source data was `rd` and the user entered `roa` then the documents would not match using query-time synonym substitution.

I suspect it would also increase the response-time significantly, and also potentially change some behaviour, so it's not a simple thing to change.

It looks like this bug exclusively affects *multi-word* synonyms, and we have relatively few of those?

missinglink:
I just looked and I couldn't find any multi-token synonyms listed in this repo, any idea which synonym is causing this?

Joxit:
Query time synonyms can be cool if we do change often our synonyms. But it's not really the case here.
And as @missinglink says, I'm a bit afraid with the response time and the CPU and IO usage that will cause the query-time synonyms...

orangejulius:
ohhh! you know what? i was testing using on of our geocode.earth client configurations. They have multi-token synonyms. so its still important to consider, but won't affect "stock" pelias

missinglink:
okay, so I have tracked down a reproducible testcase: https://gist.github.com/missinglink/8f55271dcf4f5e7e8d0712b1f2c8d742

[...]

PR #388, opened 2019-10-31
throw on multi token synonyms

missinglink:
It turns out we do have some multi-token synonyms which may have been causing us issues in elasticsearch5 (silently) and are now much more loudly noticeable in elasticsearch6.

This PR adds a new flag to the synonym parser which causes it to throw whenever it finds a multi-word synonym (and I've enabled it by default).

I've also gone through and removed all the existing multi-word synonyms from our synonyms files.

This change is inevitably going to cause a bunch of headaches for users with custom synonyms files, I'd actually like to find a solution to allow these multi-token synonyms (most likely at query time), see the linked issue for more detail.

A quick summary of why this is required is that there is an issue in elasticsearch with how it handles multi-token synonyms, if you have a substitution like this "very nice=>not bad" then what will happen is you'll end up with the following token positions:

```
position 0: [ very, not ]
position 1: [ nice, bad ]
```

that doesn't sound terrible until you consider that the queries "very nice", "very bad", "not nice" & "not bad" and now synonymous 🤷‍♂ 

> This is a story about four people named Everybody, Somebody, Anybody and Nobody. There was an important job to be done and Everybody was sure that Somebody would do it. Anybody could have done it, but Nobody did it. Somebody got angry about that, because it was Everybody’s job. Everybody thought Anybody could do it, but Nobody realized that Everybody wouldn’t do it. It ended up that Everybody blamed Somebody when Nobody did what Anybody could have.

related to https://github.com/pelias/schema/issues/381


missinglink:
Hmm.. so I'm not sure I want to merge this considering how much of a hassle it's going to cause people.

Maybe we can use this branch as a base to work on a more robust strategy that allows multi-token synonyms but somehow moves them to a query-time analysis so they can use the `synonym_graph` token filter.

missinglink:
Okay so good news is I found a solution which can replace this and allow us to keep the synonyms files as-is.

PR #389, opened 2019-10-31
fix mandatory token filter test

missinglink:
fix mandatory token filter test and refactor schema to enforce original intent.

there is an acceptance test which was supposed to ensure that *every* analyzer contains all of the mandatory token filters such as `lowercase` `unique` etc.

looking at this code today it was all messed up and not testing what it should have been, as a result, some of the analyzers were missing some of these very basic but very essential components (although I doubt the effect was noticeable).

In the process of fixing the test I've also refactored the token filters in order to achieve the original intent of the test.

### DIFF
--- a/test/settings.js
+++ b/test/settings.js
@@ -227,7 +227,10 @@ module.exports.tests.peliasZipAnalyzer = function(test, common) {
     var analyzer = settings().analysis.analyzer.peliasZip;
     t.deepEqual( analyzer.filter, [
       "lowercase",
-      "trim"
+      "icu_folding",
+      "trim",
+      "unique_only_same_position",
+      "notnull"
     ]);
     t.end();
   });
@@ -248,7 +251,10 @@ module.exports.tests.peliasUnitAnalyzer = function(test, common) {
     var analyzer = settings().analysis.analyzer.peliasUnit;
     t.deepEqual( analyzer.filter, [
       "lowercase",
-      "trim"
+      "icu_folding",
+      "trim",
+      "unique_only_same_position",
+      "notnull"
     ]);
     t.end();
   });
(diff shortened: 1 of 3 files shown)


test/settings.js
```diff
@@ -289,28 +295,29 @@ module.exports.tests.peliasStreetAnalyzer = function(test, common) {
...
     var s = settings();
     for( var analyzerName in s.analysis.analyzer ){
       var analyzer = s.analysis.analyzer[analyzerName];
       if( Array.isArray( analyzer.filter ) ){
-        analyzer.filter.forEach( function( tokenFilterName ){
-          var filterExists = s.analysis.filter.hasOwnProperty( tokenFilterName );
-          if( !filterExists && -1 < ES_INBUILT_FILTERS.indexOf( tokenFilterName ) ){
-            filterExists = true;
-          }
-          t.true( filterExists, 'token filter exists: ' + tokenFilterName );
```
orangejulius: wow I can't really figure out what the old code was intending to do, but indeed it's clearly wrong (`hasOwnProperty` on an array?)

The new code looks much better.

orangejulius:
Nice: :+1:

test/settings.js
missinglink: Agh you know what, I think it was actually doing what it was supposed to, after all, I added the original function back in.

So now there are two tests which:
- Ensure that every `token_filter` which was defined in an analyzer is actually globally defined (a sanity check)
- Ensure that a base set of mandatory token filters are present on every analyzer (an integrity check)

orangejulius: oh, that makes a lot mores sense. cool!

orangejulius:
This should be good to merge at basically any time, right?

missinglink:
Uuuummm... yes!? (he says sheepishly)

I'm assuming that the filter changes will not have a negative impact but it's still probably safer to test a build using this schema before we merge it?

missinglink:
Its a lower priority than https://github.com/pelias/schema/pull/390, so I'm happy to leave this until the ES6 stuff is merged, doing an unrelated refactor of schema while that's going on adds potential issues.

orangejulius:
They look really safe to me. We have a lot of recent builds due to the ES6 work, so I feel like this one can sneak through without too much focused attention on it.

Since it improves our tests, I also feel like merging this _before_ we merge a bunch of other improvement PRs would be helpful.

missinglink:
🔘 🖕

PR #390, opened 2019-10-31
 add flatten_graph filter to allow multi-word tokens on ES6

missinglink:
add [flatten_graph](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-flatten-graph-tokenfilter.html) token filter to ensure multi-word tokens dont produce errors during indexing.

this PR resolves https://github.com/pelias/schema/issues/381 and closes https://github.com/pelias/schema/pull/388 as an alternative.

this method was alluded to in the [elastic blogpost](https://www.elastic.co/blog/multitoken-synonyms-and-graph-queries-in-elasticsearch):

> If for some reason you really need exactly the same behavior as the old SynonymFilter then you should create a SynonymGraphFilter followed by a FlattenGraphFilter, but just remember that FlattenGraphFilter necessarily throws something (the full graph structure) away!

.. although that references a `synonym_graph` filter and not the `synonym` filter we are using... but regardless the `flatten_graph` filter serves the purpose of resolving the `illegal_argument_exception` `offsets must not go backwards` error.

the only issue here is that the operation *is* lossy, so I'm going to do a little more investigation into what exactly is lost in what situations, but I'm assuming that for any non-branching graph (such as for single-word synonyms) the filter will be a no-op.

https://www.diffchecker.com/Q1fpg2fX

### DIFF
--- a/integration/multi_token_synonyms.js
+++ b/integration/multi_token_synonyms.js
@@ -1,0 +1,49 @@
+// validate analyzer is behaving as expected
+
+const elastictest = require('elastictest');
+const schema = require('../schema');
+
+module.exports.tests = {};
+
+// simple test to cover the issue noted in:
+// https://github.com/pelias/schema/issues/381#issuecomment-548305594
+module.exports.tests.functional = function (test, common) {
+  test('functional', function (t) {
+
+    var suite = new elastictest.Suite(common.clientOpts, { schema: schema });
+    suite.action(function (done) { setTimeout(done, 500); }); // wait for es to bring some shards up
+
+    // index a document with all admin values
+    // note: this will return an error if multi-token synonyms are
+    // not supported on ES6+
+    suite.action(function (done) {
+      suite.client.index({
+        index: suite.props.index, type: 'doc',
+        id: '1', body: {
+          name: { default: 'set' },
+          phrase: { default: 'set' },
+          address_parts: {
+            name: 'set',
+            street: 'set'
+          },
+          parent: {
+            country: 'set'
+          }
+        }
+      }, done);
+    });
+
+    suite.run(t.end);
+  });
+};
+
+module.exports.all = function (tape, common) {
+
+  function test(name, testFunction) {
+    return tape('multi token synonyms: ' + name, testFunction);
+  }
+
+  for (var testCase in module.exports.tests) {
+    module.exports.tests[testCase](test, common);
+  }
+};
--- a/integration/run.js
+++ b/integration/run.js
@@ -84,7 +84,8 @@ var tests = [
   require('./bounding_box.js'),
   require('./autocomplete_street_synonym_expansion.js'),
   require('./autocomplete_directional_synonym_expansion.js'),
-  require('./autocomplete_abbreviated_street_names.js')
+  require('./autocomplete_abbreviated_street_names.js'),
+  require('./multi_token_synonyms.js')
 ];
 
 tests.map(function(t) {
--- a/settings.js
+++ b/settings.js
@@ -46,7 +46,8 @@ function generate(){
             "word_delimiter",
             "custom_admin",
             "unique_only_same_position",
-            "notnull"
+            "notnull",
+            "flatten_graph"
           ]
         },
         "peliasIndexOneEdgeGram" : {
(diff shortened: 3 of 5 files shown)


orangejulius:
:clap: :clap:

missinglink:
Okay so I think this is the correct config to use, I [tokenized the term "set"](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-analyze-http) using the following configurations:

- [master_5.6.12](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-master_5.6.12)
- [master_6.8.3](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-master_6.8.3)
- [synonym_5.6.12](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-synonym_5.6.12)
- [synonym_6.8.3](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-synonym_6.8.3)
- [synonym_graph_lenient_5.6.12](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-synonym_graph_lenient_5-6-12)
- [synonym_graph_lenient_6.8.3](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-synonym_graph_lenient_6.8.3)

The configurations are:
- `master` - the current master branch
- `synonym` - this branch which uses the `synonym` & `flatten_graph` filters
- `synonym_graph` - this branch with the `synonym` filter changed to `synonym_graph` and the `lenient` option enabled.

> note: synonym_graph without the lenient option fails on 6.8.3 with our current schema with the error [illegal_argument_exception] failed to build synonyms

Using this [test script](https://gist.github.com/missinglink/9be9893ea2771b9365588d538d2cf086#file-diff-sh) I was able to confirm that the following tokenizations are equivalent:

- `master_5.6.12` == `synonym_5.6.12`
- `master_6.8.3` == `synonym_6.8.3`

Interestingly the following tokenizations are different, meaning there are still some differences between versions which seem to be unavoidable:

- `master_5.6.12` != `master_6.8.3`
- `synonym_5.6.12` != `synonym_6.8.3`

missinglink:
This is good to merge, I'll run a full planet-build over the weekend to do some final checking.

missinglink:
Looking at the differences between `master` on the versions of ES (not directly related to this issue)...

It seems thee differences are due to how `punctuation` is being stripped (or not stripped).
I believe this is due to a change in how the token filters are applied between versions.

A hint about this behaviour can be found [buried at the bottom of the documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-tokenfilter.html) where it says:

> Elasticsearch will use the token filters preceding the synonym filter in a tokenizer chain to parse the entries in a synonym file

I don't recall hearing about this specific behaviour prior to ES6.

So what I think is happening is that ES5 and below could produce tokens which contained characters excluded by our `punctuation` token filter whereas in ES6+ the punctuation is removed, resulting in fewer tokens.

```diff
master_5.6.12/master_6.8.32,7d1
< "7."
< "7.b"
< "7.br"
< "7.bre"
< "7.r"
< "7.re"
21,25d14
< "sep."
< "sep.b"
< "sep.br"
< "sep.bre"
< "sep.e"
38,41d26
< "set."
< "set.b"
< "set.br"
< "set.bre"
```

This is a big change of behaviour which is not very well documented IMHO, it might require us to change the order in which filters get applied, possibly moving the synonym filters higher in the list.

One of the issues with this is.. if we move the synonym filter *above* lowercase (for example) then the casing of synonyms would be important, similar issues are relevant for `icu_folding` and punctuation filters.

missinglink:
Agh so on reflection this punctuation thing shouldn't be an issue after all.

The punctuation character filter is also being applied at query-time so any punctuation will be stripped from input tokens.

As a result there would have been no chance for any indexed token with punctuation remaining to match any query tokens where the punctuation has been removed.

In anything it's a no-op for query matching with the added benefit of reducing then index size slightly.

missinglink:
This is being tested and should be merged this week.

missinglink:
Okay, so it took a while but the results are finally in, all the acceptance-tests pass on ES5 with the current `master` API and this schema built on an ES5 machine, so this is now good to merge 🎉 

I've also had a play with it built using ES6 and it doesn't cause any regressions there either.

missinglink:
I rebased this against master and resolved the conflicts with https://github.com/pelias/schema/pull/389

PR #393, opened 2019-11-08
add Spanish street suffix synonyms

missinglink:
today I noticed that we were not able to match the query `Av Michoacán` against the indexed term `Avenida Michoacán`

this PR adds Spanish language synonyms for street suffixes.
consider this a  living document, please feel free to contribute/edit the file 😃


missinglink:
rebased to fix failing tests

synonyms/street_suffix.txt
```diff
@@ -127,3 +127,48 @@ bruecke, brucke, br
...
+rambla,rbla,rbl
+rambleta,rblt
+riera,ra
+ronda,rda
+rotonda,rtda
+sortida,sort,salida,slda,sort,salida,slda
+transversal,trval,trvsal,trval,trvsal
+travessera,trav,travesera,trav,travesera
+travessia,trv,trav,travesia
+via,v
```
orangejulius: Lets be sure to test this before we merge it. One character synonyms can probably have large and surprising impacts on both performance and result quality.

orangejulius:
Definitely looks like a good thing to add overall.

I made a small note about the `via,v` synonym. I think single character synonyms can have surprising impact. Are we almost at the point where we want region specific synonyms? As far as I know, that synonym would not be super useful outside Spanish speaking countries, right?

synonyms/street_suffix.txt
bboure: I have a different (and quite larger) list for this
https://gist.github.com/bboure/9adb22370bd9b4f794f0d4282662e14f

Source:
https://www.agenciatributaria.es/AEAT.internet/en_gb/Inicio/Ayuda/Tablas_auxiliares_de_domicilios__provincias__municipios____/Tabla_de_Tipos_de_Vias/Tabla_de_Tipos_de_Vias.shtml

Not sure what cost such a large list has and if it is worth it though

PR #396, opened 2019-11-20
fix(CI): update Travis-CI

missinglink:
This PR modernizes our Travis-CI config.

I had numerous problems testing Elasticsearch6 & 7 using the existing config, I've managed to backport that code to target `master`

Highlights:
- switch to `bionic` from `trusty`
- test multiple supported versions of ES (current & future)
- test multiple JVM versions (current & future long-term releases)

Probably doesn't look like many changes but took 4 hours to get this working and numerous Travis builds ☹️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
-# Once openjdk/oraclejdk dependencies are resolved, change to dist: xenial
-dist: trusty
+dist: bionic
 notifications:
   email: false
 node_js:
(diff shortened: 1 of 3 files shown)


missinglink:
One unfortunate disadvantage to having such a large build matrix is that builds take longer to run.
I think for this repo specifically it's worth it, and once we stabilize node/ES/JVM versions we can greatly reduce the matrix size.

PR #403, opened 2019-11-26
backport some features from the ES7 branch which may be ok to use now

missinglink:
backport some features from the ES7 branch which may be ok to use now

note: this includes the code which moves `word_delimiter` down one place, I'm not against removing that line as suggested in the draft PR https://github.com/pelias/schema/pull/392, I just wanted to keep this as close to a no-op as possible for now.

merging this will ensure that indices built on ES6 are more compatible with future versions, hopefully avoiding breaking changes down the line.

our new epic build matrix will catch any bugs ;)

### DIFF
--- a/mappings/partial/shape.json
+++ b/mappings/partial/shape.json
@@ -1,5 +1,3 @@
 {
-  "type": "geo_shape",
-  "tree": "quadtree",
-  "tree_levels": "20"
+  "type": "geo_shape"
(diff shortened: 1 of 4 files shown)


missinglink:
nice, so with this change, it's possible to generate indices on ES6 which are compatible with 7 while at the same time still supporting 5 🎉

PR #406, opened 2019-11-27
modernize scripts js

missinglink:
this commit was backported from the es7 branch


missinglink:
I self-reviewed this and it appears to be 100% cosmetic

orangejulius:
👍 for modernization and extracting it from a feature PR :)

PR #409, opened 2019-12-11
disable adding information to the inverted-index which isn't used

missinglink:
This PR sets the [index_options](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-options.html) param for `keyword` fields which are only used in `filter` context; so we don't ever use this information.

The following field metadata will be excluded from the index for these fields:
- freqs
- positions
- offsets

I'm opening this initially as a draft PR until we can test this out, some things to check:
- it reduces the index size sufficiently to bother
- these fields still work as expected in a `filter` context
- these fields still work as expected in an `aggregation` (such as the sources/layer agg)

### DIFF
--- a/mappings/partial/keyword.json
+++ b/mappings/partial/keyword.json
@@ -1,4 +1,5 @@
 {
   "type": "keyword",
-  "doc_values": false
+  "doc_values": false,
+  "index_options": "docs"
 }
--- a/mappings/partial/keyword_with_doc_values.json
+++ b/mappings/partial/keyword_with_doc_values.json
@@ -1,3 +1,4 @@
 {
-  "type": "keyword"
+  "type": "keyword",
+  "index_options": "docs"
 }
--- a/test/fixtures/expected.json
+++ b/test/fixtures/expected.json
@@ -715,10 +715,12 @@
     "doc": {
       "properties": {
         "source": {
-          "type": "keyword"
+          "type": "keyword",
+          "index_options": "docs"
         },
         "layer": {
-          "type": "keyword"
+          "type": "keyword",
+          "index_options": "docs"
         },
         "name": {
           "type": "object",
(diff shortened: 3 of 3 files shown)


missinglink:
I tried this out on a very small extract (just OA and OSM for Singapore) with the following results:
- index size 295M vs. 280M (index size was ~5% smaller)
- source/layer filtering still working as expected
- aggregations still working as expected
- findByID still working as expected

orangejulius:
Nice! A 5% index size reduction is definitely helpful. When I tested some other changes like this such as https://github.com/pelias/schema/pull/329, the size reduction was pretty stable across larger builds.

If this looks good on say, an Australia build, then it's probably good to merge.

missinglink:
I did a full planet build last night and the snapshot size was pretty much exactly the same as before:

```
planet-standard-es6-2019.12.05-032145-b5bdf93a-8da4-4436-b52c-d521a6e9d0d4/
389 Objects - 212.8 GB

planet-index-options-2019.12.11-153543-4abbd523-1661-4abd-9e9f-0fc55130b151/
388 Objects - 212.7 GB
```

This is not what I was expecting, and seems like it wouldn't be worth disabling these features if there is no benefit in doing so 🤷‍♂ 

One thing which I still don't understand is that when doing a `GET` request to retrieve the mapping it doesn't show the `index_options` property.

I think this is a red herring and a ~~bug~~ feature of how the `GET mapping` API works (it's the same running locally):

```bash
curl -s 'localhost:9200/pelias/_mapping/_doc' | jq '.["pelias-ci-planet-index-options-2019.12.11-153543"].mappings._doc.properties.layer'
{
  "type": "keyword"
}
```

missinglink:
opened this issue over at ES for clarification, it may just be confusion due to an omission in the docs:
https://github.com/elastic/elasticsearch/pull/50136

missinglink:
closing this as `docs` is already the default for `keyword` datatypes.
I think the 5% difference in size must have been a measurement error, possibly caused by reading the disk before the final merge operation was done.

PR #410, opened 2019-12-12
disable adding information to the inverted-index which isn't used

missinglink:
trying again with what I learned in https://github.com/pelias/schema/pull/409

this PR removes `freqs` and `positions` from fields which only ever contain a single entry.


missinglink:
ok so I'm still a little confused about the numbers here:

```
index_options
277M	/data/pelias/docker/projects/singapore/elasticsearch

master
295M	/data/pelias/docker/projects/singapore/elasticsearch
```

As I wrote in the other PR the disk seemed to be reduced to `280M` although it shouldn't have changed at all 🤷‍♂ 

So either this is a ~6.2% change or a ~1% change (280 vs 277).

I'll do one more full planet build but I'm not expecting much

missinglink:
One other thing to consider when talking about reducing disk space is the `population` and `popularity` `"null_value": 0` setting, on aggregate this consumes about 4GB per field, most of which are zeros.

missinglink:
okay so I ran this as a full planet build and it didn't reduce the size, in fact it increased by a tiny little bit.

I think this whole exercise was insightful but a bit of a waste of time, there don't seem to be any 'quick wins' for reducing the index size here.

orangejulius:
It wasn't a waste of time: now we know. :P

PR #412, opened 2019-12-13
require "analyzer", "search_analyzer" and "normalizer" to be set for all stringy fields

missinglink:
This is something I've been wanting to do for a while, it explicitly sets the `analyzer` and `search_analyzer` property for all stringy fields.

I've included a tool with this PR which can be used to list all the fields and their corresponding analyzers.

The default behaviour of elasticsearch is to default all analyzers to `standard` when not defined and to default `search_analyzer` to equal `analyzer` when it's defined.

So while it's not totally necessary to define an explicit `search_analyzer` when it's equal to the `analyzer` I have made this mandatory and covered it with tests, this ensures that it is considered when adding new fields or adapting existing ones.

I set out to try and do this as a no-op refactor (basing the `search_analyzer` settings on what's in the `defaults` for `pelias/api`), however, I discovered fairly quickly that the existing config is sub-optimal and so I've made the following changes:

```diff
 field                            type                             analyzer                  search_analyzer           normalizer
-source                           keyword                          n/a                       n/a                       {none}
-layer                            keyword                          n/a                       n/a                       {none}
-name.*                           text                             peliasIndexOneEdgeGram    peliasIndexOneEdgeGram    n/a
-phrase.*                         text                             peliasPhrase              peliasPhrase              n/a
+source                           keyword                          n/a                       n/a                       peliasKeyword
+layer                            keyword                          n/a                       n/a                       peliasKeyword
+name.*                           text                             peliasIndexOneEdgeGram    peliasQuery               n/a
+phrase.*                         text                             peliasPhrase              peliasQuery               n/a
 address_parts.name               text                             keyword                   keyword                   n/a
 address_parts.unit               text                             peliasUnit                peliasUnit                n/a
 address_parts.number             text                             peliasHousenumber         peliasHousenumber         n/a
 address_parts.street             text                             peliasStreet              peliasStreet              n/a
 address_parts.cross_street       text                             peliasStreet              peliasStreet              n/a
 address_parts.zip                text                             peliasZip                 peliasZip                 n/a
 parent.continent                 text                             peliasAdmin               peliasAdmin               n/a
 parent.continent_a               text                             peliasAdmin               peliasAdmin               n/a
-parent.continent_id              keyword                          n/a                       n/a                       {none}
+parent.continent_id              keyword                          n/a                       n/a                       peliasKeyword
 parent.ocean                     text                             peliasAdmin               peliasAdmin               n/a
 parent.ocean_a                   text                             peliasAdmin               peliasAdmin               n/a
[...]

PR #414, opened 2019-12-13
Explicitly set analyzers

missinglink:
cherry-picked from https://github.com/pelias/schema/pull/412 and based on https://github.com/pelias/schema/pull/413. [view diff](https://github.com/pelias/schema/compare/list_analyzers...explicitly_set_analyzers)

This is something I've been wanting to do for a while, it explicitly sets the `analyzer` and `search_analyzer` property for all `text` fields.

The default behaviour of elasticsearch is to default all analyzers to `standard` when not otherwise defined..
... and to default the `search_analyzer` property to equal `analyzer`.

So while it's not totally necessary to define an explicit `search_analyzer` when it's equal to the `analyzer` I have made this mandatory and covered it with tests, this ensures that it is considered when adding new fields or adapting existing ones.

This is hopefully a no-op refactor (basing the `search_analyzer` settings on what's in the `defaults` for `pelias/api`).
It will benefit any queries where the `analyzer` was not set on the query for whatever reason.

```diff
 field                            type                             analyzer                  search_analyzer           normalizer
< name.*                           text                             peliasIndexOneEdgeGram    peliasIndexOneEdgeGram    n/a
< phrase.*                         text                             peliasPhrase              peliasPhrase              n/a
---
> name.*                           text                             peliasIndexOneEdgeGram    peliasQuery               n/a
> phrase.*                         text                             peliasPhrase              peliasQuery               n/a
```

### DIFF
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -26,26 +26,32 @@ var schema = {
         name: {
           type: 'text',
           analyzer: 'keyword',
+          search_analyzer: 'keyword'
         },
         unit: {
           type: 'text',
           analyzer: 'peliasUnit',
+          search_analyzer: 'peliasUnit'
         },
         number: {
           type: 'text',
           analyzer: 'peliasHousenumber',
+          search_analyzer: 'peliasHousenumber'
         },
         street: {
           type: 'text',
           analyzer: 'peliasStreet',
+          search_analyzer: 'peliasStreet'
         },
         cross_street: {
           type: 'text',
           analyzer: 'peliasStreet',
+          search_analyzer: 'peliasStreet'
         },
         zip: {
           type: 'text',
           analyzer: 'peliasZip',
+          search_analyzer: 'peliasZip'
         },
       }
     },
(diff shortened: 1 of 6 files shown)


missinglink:
I know there are some queries [such as this](https://github.com/pelias/query/blob/8fcb336077ff09fe2a5b47436924b86e6252ce3a/test/fixtures/addressesUsingIdsQuery/with_layers_and_filters.json) where the `analyzer` is not being set.

It's hopefully not a big deal right now that query doesn't target the `name.*` and `phrase.*` fields which require a different `search_analyzer` from `analyzer`.

In the future, I'd like to explore having a new `search_analyzer` for `address_parts.street` since right now it's doing a bunch of synonym substitutions at query-time which can likely be avoided.

missinglink:
Also worth noting the `parent.{placetype}.ngram` fields were previously using and `ngram` analysis at query-time which is just plain wrong.

missinglink:
This looks good to me.
It should provide an immediate performance benefit for queries targeting the admin ngrams fields as well as any queries which were failing to correctly use the `peliasQuery` analyzer.

PR #416, opened 2019-12-13
add 'trim' to the 'peliasKeywordNormalizer' filters

missinglink:
This PR is based on https://github.com/pelias/schema/pull/415. [view diff](https://github.com/pelias/schema/compare/normalizers...normalizers_es6)

A nice-to-have but not a biggie, this PR ensures that normalized keyword fields are whitespace trimmed.

AFAIK to the `trim` filter not being supported in a normalizer from Lucene 7.3 (Elastic v6.4) so I suspect the tests will fail for ES5


Joxit:
I think Travis agrees :smile:

PR #394, opened 2019-11-19
elasticsearch 7

missinglink:
es7 compatibility (breaks compatibility with es5)

notable changes:
- changes inherited from https://github.com/pelias/schema/pull/385
- ~~moved 'word delimiter' filter below synonyms filter to avoid `message [failed to create index], failure [IllegalArgumentException[Token filter [word_delimiter] cannot be used to parse synonyms]]` error~~ (backported to master already)
- ~~changed the `_type` from "doc" to "_doc" (as required in ES7)~~ superseded by https://github.com/pelias/config/pull/119
- added a helper to deal with the new `hits.total` format

todo:
- [x] decide on whether we remove `word_delimiter` ~~or move it down below the synonyms~~
- [x] squash the commits and/or add a breaking change commit message
- [x] update `pelias/config` to set the default `schema.typeName` to `_doc` (done in https://github.com/pelias/pelias/issues/831)
- [x] ~~update `pelias/model` to use the `_doc` type instead of `doc`~~ (see above)
- [x] ~~update all the importers to use the updated `pelias/model` version~~ (also no longer relevant)

PR #425, opened 2020-01-08
chore(CI): remove oraclejdk8 from test matrix

orangejulius:
_Extracted from https://github.com/pelias/schema/pull/394_

As Elasticsearch 7 will come with JDK 11 out of the box, we will eventually not care about JDK 8 support.

Once we are ready to drop that support, we can merge this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,9 @@ node_js:
   - 12
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


orangejulius:
Just noticed this old PR. We are pretty much only on ES7 these days, and I don't believe testing old versions of the JDK are worth it anymore.

Issue #428, opened 2020-01-10
Issue with creating schema


srflaxu40:
Any reason for this error using the latest pelias schema repo master and ES 5.5?

```
[0;33m--------------
 create index 
--------------[0m

[class_cast_exception] java.lang.String cannot be cast to java.util.Map ```

missinglink:
We actually dropped support for ES5 this week (as ES7 support is coming shortly)
Please try again using the latest version of ES6

missinglink:
related: https://github.com/pelias/pelias/issues/719#issuecomment-559063037

missinglink:
Please do a `pelias compose pull` and try again, it should still fail but with a descriptive error message :)

PR #429, opened 2020-01-10
add check_version.js script

missinglink:
this PR adds a new script which checks the ES version of the server before attempting to create the schema.
it provides a more user-friendly error when encountering an unsupported version:

Screenshot 2020-01-10 at 11 56 47

inspired by https://github.com/pelias/schema/issues/428

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "elasticsearch": "^16.0.0",
     "lodash": "^4.17.15",
     "pelias-config": "^4.5.0",
-    "pelias-logger": "^1.3.0"
+    "pelias-logger": "^1.3.0",
+    "semver": "^7.1.1"
   },
   "devDependencies": {
     "difflet": "^1.0.1",
--- a/scripts/check_plugins.js
+++ b/scripts/check_plugins.js
@@ -5,14 +5,10 @@ const client = new es.Client(config.esclient);
 const cli = require('./cli');
 
 // mandatory plugins
-const plugins = [ 'analysis-icu' ];
+const required = ['analysis-icu'];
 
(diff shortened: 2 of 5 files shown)


missinglink:
@orangejulius could you please double-check the version semver is correct for the work you've been doing recently?
reference: https://github.com/npm/node-semver#readme

orangejulius:
This is great. Definitely a better experience.

Do you think we should make the minor and patch versions more lenient? We haven't tested on, for example, Elasticsearch 6.1, but things would probably work.

Or as another example, if someone was stuck on Elasticsearch 7.5.0 instead of 7.5.1, should we really refuse to create the index?

missinglink:
Yeah good point, I just set it to the versions which I could personally vouch for 🤷‍♂.

I'd like to merge this soon regardless as we had two bug reports for this issue today.

orangejulius:
Should we also add a check for the `apiVersion` configured through `pelias.json`? In https://github.com/pelias/docker/issues/163, the user had explicitly set version `5.6` in their own `pelias.json`.

Even if the Elasticsearch server they are running is a compatible version, this would cause issues since it would generate a create index request that causes an error.

missinglink:
We could do, but I'm happy to add that in a subsequent PR.

As per our offline discussion I think we can reduce the 6 version to `6.5+`

missinglink:
commands I was using to test:

```bash
docker run --rm --name elastic-test -p 9200:9200 pelias/elasticsearch:5.6.12

docker run --rm --name elastic-test -p 9200:9200 pelias/elasticsearch:6.8.5
```

Issue #408, opened 2019-12-11
discuss: similarity algorithms


missinglink:
For years now we've been fighting the `TF/IDF` algorithm and more recently we've changed to the `BM25` similarity algo which is much better for short texts like ours but it's still not perfect.

The cool thing about [BM25](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-similarity.html#bm25) (and [other similarity algos](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-similarity.html)) is that they have some tunable parameters, albeit considered 'expert settings'.

One setting that interests me, in particular, is the [k1 value](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-similarity.html#bm25) which "Controls non-linear term frequency normalization (saturation).".

The default settings for `BM25` are `k1=1.2` and `b=0.75`, which are really nice settings for general use of elasticsearch, they work well for short fields like titles as well as for large fields like indexing a whole chapter of a book.

For geocoding specifically we almost exclusively deal with short strings (<50 chars).
I also personally feel that Term Frequencies are much less relevant for geocoding because they can cause [issues like this](https://github.com/pelias/openstreetmap/issues/507)

I'd like to open this up to @pelias/contributors to discuss introducing our own custom similarity configuration (or multiple if required).
In particular, I would like to investigate the effects of setting `k1=0` (or very very low).

Thoughts?

missinglink:
These articles are great, especially in the second one:
- https://www.elastic.co/blog/practical-bm25-part-1-how-shards-affect-relevance-scoring-in-elasticsearch
- https://www.elastic.co/blog/practical-bm25-part-2-the-bm25-algorithm-and-its-variables

missinglink:
Some examples of existing TF/IDF scoring I don't like (which I'll update over time):

| search term | indexed term | description |
|---|---|---|
| "mcdonalds" | "mcdonalds (mcdonalds cafe)" | term frequency causes duplicate terms to rank higher |
| "ave" | "avenue" (ngrams) | IDF of ngrams saturated by prefixes of other terms |

shuaibiyy:
Hi @missinglink we've found IDF to be useful in cases where documents contain words identifying their category. Let's say we have a query `hotel park` where the user is searching for a hotel called `Park Inn`. Some hotels might prefix their name with the word `Hotel`, while some might not. Assuming these are some of the documents retrieved:
* Ibis Hotel
[...]

PR #430, opened 2020-01-10
add peliasDefaultSimilarity

missinglink:
This PR adds a `peliasDefaultSimilarity` similarity module and configures it with the defaults for `BM25` (making this a no-op).

```bash
curl -s 'localhost:9200/_settings' | jq '.pelias.settings.index.similarity'
{
  "peliasDefaultSimilarity": {
    "type": "BM25",
    "b": "0.75",
    "k1": "1.2"
  }
}
```

The advantage of using a custom similarity over using the default (which is also `BM25`) is that the values of `k1` and `b` can be configured.

One way of changing the `BM25` settings is via editing `pelias.json` **before creating the index** as such:

```javascript
{
  "elasticsearch": {
    "settings": {
      "index": {
        "similarity": {
          "peliasDefaultSimilarity": {
            "k1": 0,
            "b": 0
          }
        }
      }
    }
  }
}
```

```bash
curl -s 'localhost:9200/_settings' | jq '.pelias.settings.index.similarity'
{
  "peliasDefaultSimilarity": {
    "type": "BM25",
    "b": "0",
    "k1": "0"
  }
}
```

Alternatively it seems possible to avoid reindexing an existing index by simply [closing the index, changing the setting and reopening it](https://github.com/elastic/elasticsearch/issues/6727#issuecomment-573052670):

```bash
curl -sX POST 'localhost:9200/pelias/_close'

curl -sX PUT -H 'Content-Type: application/json' 'localhost:9200/pelias/_settings' -d '{
  "index": {
    "similarity": {
      "peliasDefaultSimilarity": {
        "type": "BM25",
        "k1": 0,
        "b": 0
      }
    }
  }
}'

curl -sX POST 'localhost:9200/pelias/_open'
```

> note: it's not safe to change between disparate similarity modules, changing these two BM25 settings is ok, but changing anything which modifies what gets stored in the index will obviously require a full reindex, see the link above for more info.

This PR is, by default, a no-op, but allows us to experiment with the `k1` and `b` settings.

connected to https://github.com/pelias/schema/issues/408


missinglink:
I also made some unrelated changes to https://github.com/pelias/schema/pull/430/files#diff-2d78dada12f5fa4f3183bc45f30f590c, @orangejulius could you please spot check this for me?

PR #436, opened 2020-03-26
fix(travis): loosen ES api version to only target major version number

missinglink:
as per https://github.com/pelias/schema/issues/435, the elasticsearch library only allows targeting the latest version (such as `7.6`) or `7.x`.

this means that if we have an env var such as `ES_VERSION=7.5.1` in our Travis tests then the generated version `7.5` is invalid once a new version is available, this PR solves this.

### DIFF
--- a/scripts/setup_ci.sh
+++ b/scripts/setup_ci.sh
@@ -37,7 +37,7 @@ v=( ${ES_VERSION//./ } ) # split version number on '.'
 # generate a pelias.json config
 PELIAS_CONFIG=$(
   jq -n \
-    --arg apiVersion "${v[0]}.${v[1]}" \
+    --arg apiVersion "${v[0]}.x" \
     '{
       esclient: {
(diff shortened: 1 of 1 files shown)


missinglink:
ugh, https://travis-ci.org/github/pelias/schema/jobs/667151454?utm_medium=notification&utm_source=github_status I hate how inconsistent ES configs can be 😡

missinglink:
I can't be bothered trying to IF/ELSE this, replacing this PR with https://github.com/pelias/schema/pull/437

orangejulius:
Bummer. I'm guessing they removed `6.x` because they are confident that there won't be any more Elasticsearch 6 minor releases.

missinglink:
Honestly I think the lack of thought and reckless commits are because they don't care about the legacy version of the js client since there is a newer one.

Maybe we should just bite the bullet and migrate before they give up on it entirely?

Issue #443, opened 2020-06-01
Synonyms for grey vs gray


orangejulius:
We should handle difference is spelling of `grey` vs `gray` with a synoynm.

We probably cannot rely on users to type the one that matches in a streetname, venue name, etc. In fact, over in https://github.com/pelias/api/pull/1432#issuecomment-636944229 I found a case where OSM data uses one spelling for a street, and data from OA uses the other.

What are other English language words that have common spelling variants like this? We might want to handle them all.

PR #407, opened 2019-12-05
Remove peliasQueryFullToken analyzer

Joxit:
Since we merged our beloved pelias parser, this analyser is no more used :wave:.

Issue #301, opened 2018-06-08
Synonyms for french addresses


adefarge:
Hi,
French addresses don't work when using the search feature with abbreviations.
For example:
* `boulevard saint-martin` works
* `bd saint-martin` doesn't work
* `boulevard saint martin` doesn't work either

Ideally, we would like to find the address with the input `bd st martin`.

I see two problems here:
1. The synonyms in `custom_street.txt` don't work because the `peliasStreetTokenizer` doesn't split on whitespace. But I see you are fixing this in #291 
2. `saint martin` doesn't match `saint-martin` because of the hyphen. This can be solved by adding `-` to the name and street tokenizer.
3. Libpostal sometimes parses martin as the neighbourhood or the city instead of the street. But that is another issue.

A more complete list of localized synonyms is also needed but that is less a problem because we can just add our own in our local install of pelias.

What do you think ?

missinglink:
This all sounds good.

I am always nervous about changing the tokenizers because it changes the system so much.

But, yes; in this case, I think this change would be for the best.

In order to do this change safely we will need to run a build and test to ensure we haven't missed any edge cases.

I would suggest we start a new branch which is clean, from there we make the minimal amount of changes required to improve `peliasStreetTokenizer` and then go from there.

orangejulius:
On that note, I'm going to split out just the change to tokenize on whitespace from #291, so that we can test that change individually as well.

missinglink:
This would also be a good opportunity to start putting together a French test suite.

adefarge:
Okay,
For the test suite, were you talking about acceptance tests, or tests in the schema module ?
I can try to put together a few test cases for french addresses, but I will need to know what kind of data you need (just the name of a street, a full address, an OSM link ?).

I suppose you'll want to do the change to the tokenizer yourself ? Please tell me if I can help with anything.

loicortola:
@missinglink, following up on this topic, this might be a game-changer for the adoption of Pelias in the french community :) .

@orangejulius any update on the split of #291 ?

@adefarge is available to work for a few weeks on the project, are you guys open to getting organized so that we take advantage of the momentum to safely make and test those schema changes?

If yes, please let us know how :)

Cheers!

orangejulius:
@adefarge, we are looking for primarily french acceptance-tests. we need your local knowledge :)  of course we will not say no to unit tests

[...]

PR #446, opened 2020-06-12
update street synonyms

missinglink:
This WIP PR is to improve the synonyms specifically for USA streets.
Motivated by the following matching failures:

```
Boul === Boulevard
Crt === Court
Ter === Terrace
```

I've added the USPS synonyms list which should cover a large amount of USA street names.

I'd like to see the effect of this on a full-planet build before deciding if this is something we might consider doing more of, for other languages and classes of synonyms (such as Gray<>Grey and 1st<>First).


missinglink:
I'm still not sure how comfortable I am with some of these synonyms which come from libpostal:

eg:
```
l,lane
road,raod
```

missinglink:
rebased on top of https://github.com/pelias/schema/pull/447

synonyms/street_synonyms_en.txt
```diff
@@ -0,0 +1,379 @@
...
+retreat, rt, rtt
+return, rtn
+ridge, rdge, rdg
+ridges, rdgs
+ridgeway, rgwy, rdgwy, ridge way
+right of way, rowy, rightofway, rofw, row, r o w, r of w
+rise, ri
+riverway, rvwy
+riviera, rvra
+road, rd, ro, r, roa, raod
```
orangejulius, line 283: Yeah, maybe we leave out any single character synonyms for now?

And definitely typos should not be synonyms.

settings.js
```diff
@@ -68,7 +72,9 @@ function generate(){
             "icu_folding",
             "trim",
             "custom_name",
-            "street_suffix",
+            "street_synonyms_en",
+            "street_synonyms_usps",
```
Joxit, line 76: How about having something with a regex like `street_synonyms_*` and will fetch all the files ?
This may help for #393 or if we need to add custom street_synonyms by lang.

synonyms/street_synonyms_en.txt
```diff
@@ -0,0 +1,379 @@
...
+basin, basn, bsn
+bay, by
+bayou, byu, bayoo
+beach, baech, bch, beech
+belt, blt
+bend, bnd
+block, blk, blck
+bluff, blf, bluf, bluffs, blfs
+boardwalk, bwk, bwlk, board walk
+boulevard, blvd, bd, bde, blv, bl, blvde, blvrd, boulavard, boul, boulv, bvd, boulevarde
```
Joxit, line 28: This one will help #301 and the example `bd saint-martin`

settings.js
missinglink: Yeah good point, we should rethink these a bit and make them easier to customise.

Right now we have fairly limited synonyms coverage (ie. 'just the basics').
I'd like to explore what would happen if we added significantly more synonyms, I suspect that the impact would be very positive for search quality and hopefully doesn't affect performance much.

@Joxit could you please tell me a bit about how you're adding custom synonyms for your builds?

I'm guessing you're adding a bunch of French-specific and some other ones which you include in the `custom_street.txt` file?

Pending successful testing of this I would next like to expand on French, German and Spanish synonyms.

Joxit: This is a bit annoying but better since #375 and #407 

I have a fork [joxit/pelias-schema](https://github.com/Joxit/pelias-schema), when I start a new build, I fetch the upstream and add [this commit](https://github.com/Joxit/pelias-schema/commit/514fb027cb5dd3b2be5ad1f4da62db0f24f20951).

For the record, I added `cutom_street` in `peliasPhrase` and `peliasQuery` in `build-2019-09` (before `pelias/parser` release). `peliasQuery` is used for autocomplete where we can find streets names... And `peliasPhrase` because `street_suffix` is also present.

I add my synonyms in `synonyms/custom_street.txt` 
```
bd, bld, blvd, boul, bvd, boulevard #only bld is missing in this PR
ave, av, avenue # OK with this PR
saint, st, street # missing
sainte, ste, suite # missing
```
This list was created from our common uses (when we test the pelias). I have always been afraid of slowing down queries with too many synonyms :sweat_smile:.

synonyms/street_synonyms_en.txt
missinglink: I've added another PR to remove a bunch of single char synonyms (although I'm not sure if they are such a big issue?)
Also removed some synonyms from the `libpostal` list which appeared to only be there for the purpose of spelling correction.

missinglink: There is a recent issue opened specifically suggesting adding a synonym for typos https://github.com/pelias/schema/issues/443

Let's chat about this some more, for now I'm going to remove them.

missinglink:
Okay, I'm going to run this as a full-planet build and see what effect it has on the acceptance tests.

I'm particularly interested to see if adding loads of synonyms is feasible from quality/perf perspective 🤷

settings.js
missinglink: Do you possibly know of an official source which publishes a French equivalent of this?
https://pe.usps.com/text/pub28/28apc_002.htm

missinglink: I've opened https://github.com/pelias/schema/pull/449 to hopefully clean up your build process a little.

Ideally you shouldn't have to do any `synonyms` work in `peliasQuery`, I think you can remove that and you'll get a small perf benefit?

blackmad:
Thanks for working on this, Peter!

settings.js
Joxit: I found this one : https://blog.bureaudeposte.net/rediger-une-adresse-aux-normes-postales/ for abbreviations

Name| abbreviation
-- | --
Allée | ALL
Avenue | AV
Boulevard | BD
Carrefour | CARR
Chaussée | CHS
Chemin | CHEM
Cours | CRS
Faubourg | FG
Immeuble | IMM
Impasse | IMP
Lieu-dit | LD
Lotissement | LOT
Montée | MTE
Passage | PAS
Place | PL
Résidence | RES
Route | RTE
Ruelle | RLE

Joxit: Next time I will check differences between synonyms in `peliasQuery` and #449 

Oh, and saint/sainte aren't for street suffix/prefix but a workaround for street names... :sweat_smile:

missinglink:
Looks pretty good to me 🎉 
https://www.diffchecker.com/qlBPvxAO

missinglink:
I've removed all the multi-word synonyms in https://github.com/pelias/schema/pull/446/commits/5852086ac1419fa00e8c50fd259148fd891ce8cd, these are potentially dangerous, especially when used at query time (such as is the case right now for `peliasStreet` fields)

orangejulius:
One additional synonym we might want to add here is `st/saint`. This would fix https://github.com/pelias/pelias/issues/737

We can also do it as a followup PR if that's easier, but let's make sure not to forget :)

orangejulius:
This PR has been replaced by #453

Issue #456, opened 2020-07-31
Multi token synonyms appear broken by synonyms upgrade


orangejulius:
Since #453, it looks like any Pelias users who wish to define custom multi-token synonyms are out of luck.

Here's a small test script to demonstrate the change in behavior:

```sh                                                                                                                                                                                                                                                                                                                                                                                                                                                          
git checkout v5.6.0 # after major synonyms upgrade (https://github.com/pelias/schema/pull/453)                                                                                                                                                 
                                                                                                                                                                                                                                               
# clear pelias index for clean slate                                                                                                                                                                                                           
node scripts/drop_index.js --force-yes &> /dev/null                                                                                                                                                                                            
                                                                                                                                                                                                                                               
# set up a custom multi token synonym                                                                                                                                                                                                          
echo "aaaa bbbb cccc dddd, abcd" >> synonyms/custom_street.txt                                                                                                                                                                                 
                                                                                                                                                                                                                                               
node scripts/create_index.js &> /dev/null                                                                                                                                                                                                      
                                                                                                                                                                                                                                               
curl -s "localhost:9200/pelias/_analyze" -H 'Content-Type: application/json' \                                                                                                                                                                 
    -d '{ "text": "aaaa bbbb cccc dddd", "analyzer": "peliasStreet" }' | jq '.tokens[] | {token}'                                                                                                                                              
                                                                                                                                                                                                                                               
git checkout v5.5.1 # before major synonyms upgrade (https://github.com/pelias/schema/pull/453)                                                                                                                                                
                                                                                                                                                                                                                                               
# clear pelias index for clean slate                                                                                                                                                                                                           
node scripts/drop_index.js --force-yes &> /dev/null                                                                                                                                                                                            
                                                                                                                                                                                                                                               
# set up a custom multi token synonym                                                                                                                                                                                                          
echo "aaaa bbbb cccc dddd, abcd" >> synonyms/custom_street.txt                                                                                                                                                                                 
                                                                                                                                                                                                                                               
node scripts/create_index.js &> /dev/null                                                                                                                                                                                                      
                                                                                                                                                                                                                                               
curl -s "localhost:9200/pelias/_analyze" -H 'Content-Type: application/json' \                                                                                                                                                                 
    -d '{ "text": "aaaa bbbb cccc dddd", "analyzer": "peliasStreet" }' | jq '.tokens[] | {token}'
```

On my machine, this prints only the 4 input tokens on the latest version of schema:
```
{
  "token": "aaaa"
[...]

PR #457, opened 2020-08-04
re-enable support for custom multi-word synonyms

missinglink:
As discussed in https://github.com/pelias/schema/issues/456, the work in https://github.com/pelias/schema/pull/453 had the unexpected consequences of dropping support for multi-word custom synonyms.

My general guidance here is that multi-word synonyms are poorly supported by lucene/elasticsearch and so should be avoided where possible, great care should be taken to ensure they are compatible with the `match_phrase` queries used by Pelias.

> this new elasticsearch doc explains the issues with multi-word synonyms with accompanying graphics:
> https://www.elastic.co/guide/en/elasticsearch/reference/7.6/token-graphs.html

Where possible I'd recommend using 'aliases' ie. `doc.setNameAlias()` instead, this is a reliable method of achieving the same thing, although it's far less convenient because it's on a per-record basis.

So.. having said all that.. this PR re-enables support for multi-word synonyms (in custom synonyms files only) in order to avoid breaking backwards compatibility.

In order to do this I had to move the custom multi-word synonyms outside the multiplexer, apparently multiplexers emit tokens one-by-one to each of their branches, preventing the ability to 'look-ahead' as required by any multi-term analysis within a branch.

Screenshot 2020-08-04 at 10 21 15

resolves https://github.com/pelias/schema/issues/456


synonyms/linter.js
```diff
@@ -74,6 +75,12 @@ function multiWordCheck(line, logprefix, tokens) {
   });
 }
 
+function tokenReplacementCheck(line, logprefix) {
```
missinglink, line 78: I've also added this additional linter warning.

Technically this syntax is actually still supported (the linter doesn't rewrite the configs) but I would like to heavily discourage its use.

The reason for this is an entry such as `road => rd` would only enter `[ "rd" ]` in the index, so when a user queries for `[ "road" ]` (or any ngram of) it will not match.

Advanced users may opt to write it as `road => road, rd` to overcome this issue but this is equivalent to `road, rd`.

I think we will probably completely drop support for `=>` at some point in the future because it's very easy to make make mistakes with.

missinglink:
I've added https://github.com/pelias/schema/pull/457/commits/1baeadfe9529e8ede8207c8a688ee5459efe3c48 to address an issue where the linter is was using `/\s/` instead of `[\\s/\\\\-]+` to determine which tokens were multi-word.

As a result I've had to remove a few hyphenated synonyms from the canonical synonym lists.

synonyms/linter.js
orangejulius: I was thinking it might actually be useful to use `=>` sometimes. For example, if we have a single letter abbreviation that we want to ensure is not being applied too often.

`c, carrer` would replace every instance of `c` with `carrer`. However, `carrer => career, c` would not.

As you mentioned we would probably recommend that all tokens on the left hand side of the replacement also appear on the right, to avoid issues with autocomplete jitter.

orangejulius:
Nice. Does it make sense to add an integration test for multi-word synonyms?

PR #459, opened 2020-08-07
Add support for custom sources in hierarchy

Joxit:
This PR adds new `{placetype}_source` properties and should be merged before pelias/model#128 and pelias/api#1481.

related: pelias/model#128 pelias/api#1481

### DIFF
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -71,76 +71,91 @@ var schema = {
         continent: admin,
         continent_a: admin,
         continent_id: keyword,
+        continent_source: keyword,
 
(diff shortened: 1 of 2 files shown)


missinglink:
LGTM :+1:

missinglink:
nice catch on the `postalcode_id`, didn't realise that was missing...

orangejulius:
> nice catch on the postalcode_id, didn't realise that was missing...

I think it was just missing the comma at the end, being the last property defined. But we don't need it anyway :)

PR #460, opened 2020-08-11
remove support for creating indices using Elasticsearch 6.x

missinglink:
BREAKING CHANGE: creating *new* Pelias indices using elasticsearch 6 is no longer supported

see https://github.com/pelias/pelias/issues/881

Attempting to run `create_index.js` against a `pelias/elasticsearch:6.8.5` server results in the following error:

```bash
 node scripts/create_index.js

--------------
 create index
--------------

unsupported elasticsearch version. try: >=7.4.2
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ node_js:
   - 12
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


orangejulius:
Sounds good! :shipit:

PR #461, opened 2020-08-11
remove alpha ordinals from both queries and docs

blackmad:
- sometimes a street is in OSM as "twenty fifth street", which means if
  a user searches for it as "25th st", they won't find it
- sometimes a user will search for "twenty fifth street", but it's in
  OSM as "25th st"

this change takes care of both of those by synonymizing the "alpha
ordinal" to the bare number (25) in both cases above.

It does not lead to an index explosion because most streets are in OSM
as "NNth St", and this is not a bi-directional synonym.

This doesn't help with autocomplete, a user searching for "twent...."
will not find much of anything.

I tried very hard to get this to work with a multplexer filter of the
multiword + single-word filters and ... I just could not get it to work,
it would seemingly ignore one of the sub filters I specified no matter
what I did.

replaces https://github.com/pelias/schema/pull/442


orangejulius:
Full planet build with this PR has been kicked off, should be ready for testing tomorrow.

orangejulius:
So I finally got a chance to test this out and it looks like it now causes matches between lots of different housenumbers. For example, the query [](https://pelias.github.io/compare/#/v1/autocomplete?text=H%C3%A4meenkatu+1) now matches not only for housenumber `1` but `21`, `31`, `41`, etc.
Screenshot_2020-09-01_19-25-14

I'm thinking this is because `first`, `twenty first`, `thirty first`, etc all share tokens. So we'll probably have to figure out a slightly more sophisticated way to do this. Or maybe we can drop the multi-token synonyms?

blackmad:
hmmm, weird, why is it modifying the housenumbers at all. Let me look/think again, but I'm also fine with dropping the multi-words for now in the short term

orangejulius:
Ah yes important point and one we all still get confused by: the synonyms are being replaced in the `name.default` field. After https://github.com/pelias/api/pull/1469 which was merged a few weeks back, that's the only field (besides `phrase.default` which is very similar) that autocomplete looks at.

So it might be that we can restrict these synonyms to `address_parts.street`, and then the search endpoint can benefit, with autocomplete being unaffected.

PR #464, opened 2020-09-03
add a way to run integration tests locally + dockerized, for macs!

blackmad:
I was having trouble running the integration tests on a mac so I added this.


missinglink:
I'm also on mac, what issue were you having when running the tests?

I usually just run this in one terminal window:
https://github.com/pelias/schema#running-elasticsearch-in-docker-for-testing-purposes

And then this in another terminal:
https://github.com/pelias/schema#running-integration-tests

blackmad:
that elastic docker image doesn't have icu installed

orangejulius:
We use Pelias [Elasticsearch](https://hub.docker.com/r/pelias/elasticsearch/tags) docker images to solve that problem (and others). They come from [within the pelias/docker project](https://github.com/pelias/docker/tree/master/images/elasticsearch) if you want to build a new version.

PR #471, opened 2021-01-12
disable term frequencies for admin abbreviation fields

missinglink:
This PR changes the way `parent.*_a` fields are indexed so that the term frequencies are not stored.
the effect of this change is that indexing multiple terms in the field (ie. what we call 'aliases') will have no adverse effect on scoring.

Merging this work will allow us to follow up with a alpha3<>alpha2 country code mapping (for `parent.country_a`) without worrying about the additional tokens adversely affecting scoring.

related:
- https://github.com/pelias/schema/pull/470
- https://github.com/pelias/schema/pull/469
- https://github.com/pelias/schema/issues/408

### DIFF
--- a/integration/admin_abbreviations.js
+++ b/integration/admin_abbreviations.js
@@ -1,0 +1,126 @@
+/**
+ * this test suite ensures that 'admin_abbreviation' fields
+ * are indexed in a way that indexing multiple tokens in the
+ * same field does not affect scoring even when the tokens
+ * are duplicates or prefixes of each other.
+*/
+
+const elastictest = require('elastictest');
+const config = require('pelias-config').generate();
+const getTotalHits = require('./_hits_total_helper');
+
+module.exports.tests = {};
+module.exports.tests.functional = function (test, common) {
+  test('functional', function (t) {
+
+    var suite = new elastictest.Suite(common.clientOpts, common.create);
+    suite.action(done => setTimeout(done, 500)); // wait for es to bring some shards up
+
+    // index document 1 with country_a='NZL'
+    suite.action(done => {
+      suite.client.index({
+        index: suite.props.index,
+        type: config.schema.typeName,
+        id: '1',
+        body: {
+          parent: {
+            country_a: ['NZL']
+          }
+        }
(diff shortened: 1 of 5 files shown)


orangejulius:
So, looking at the [index_options](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-options.html) documentation, this is a pretty simple change :)

Code-wise, it looks good, the `admin_abbreviation` definition being built on the `admin` one is especially nice.

What do you think next steps should be to test this?

As I recall, we care most about how this affects autocomplete queries for _addresses_, right? In that case there's probably no getting around a full planet build.

Alternatively if we care mostly about queries like `city, country_code` (`Amsterdam, NL` vs `Amsterdam, NLD` for example), then maybe we can get away with a WOF only build?

missinglink:
It should only have an effect on documents which have multiple terms indexed in an abbreviation field.
AFAIK that doesn't currently happen anywhere at the moment so I'm expecting it to be a no-op.

In the case where we do actually have multiple terms indexed in an abbreviation field it would change the scoring slightly, in a positive way.

Essentially it should be a NOOP

orangejulius:
Ok, good news is that the changes in #472 _do_ fix the country code issues we've been going after.

Bad news is they break the structured geocoding endpoint, since they run phrase queries across fields that include the abbreviation fields, like this:

```json
{
    "multi_match": {
        "query": "United States",
        "type": "phrase",
        "fields": [
            "parent.dependency",
            "parent.dependency_a"
        ]
    }
}
```

So we have to find either some query or schema changes to handle that.

Here's a question: if we are using synonyms to handle the 2/3 letter country codes, do we need to disable the field length at all? My recollection is that if we tell elasticsearch to index the data "MEX", for example, and then there is a synonym "MEX,MX", the field length will still be 1. whereas if we indexed the data "MEX MX", that wouldn't be the case.

missinglink:
I'm going to abandon this PR, I seems that some queries I hadn't considered were using `phrase` type queries against these fields, as long as that functionality is required this solution is unviable.

PR #473, opened 2021-02-09
enable country code synonym file for all admin fields

missinglink:
This PR is essentially https://github.com/pelias/schema/pull/472 without https://github.com/pelias/schema/pull/471

Some of the issues I mentioned in https://github.com/pelias/schema/pull/472 are still a concern, but we can address them independently of this PR.

closes https://github.com/pelias/schema/pull/472
closed https://github.com/pelias/schema/pull/471


missinglink:
@orangejulius could you please give this a final check before we merge it?

synonyms/country_codes/generate.js
```diff
@@ -0,0 +1,27 @@
...
+const path = require('path');
+const iso3166 = require('iso3166-1');
+const filename = path.join(__dirname, 'iso3166.txt');
+const mapper = (row) => {
+  const columns = [row.alpha2, row.alpha3];
+
+  // comment-out synonyms where alpha2 is a prefix of alpha3
+  // if (row.alpha3.startsWith(row.alpha2)){
+  //   columns.unshift('#')
+  // }
```
orangejulius, line 11: Don't we want this code? :)

missinglink:
ok ok @orangejulius this is going to make you happy, I had another think about our discussion earlier and I think you're right.

I realised there is a compelling reason why we might want to exclude synonyms where the `alpha2` is a prefix of the `alpha3`, namely the issues of the ugly cross-contamination of synonyms and mess of analysers mentioned in https://github.com/pelias/schema/pull/472

In particular just not having `STP,ST` applied to the `peliasOneEdgeGram` analyser alone is a very compelling reason.

It does come with some tradeoffs however, which are illustrated nicely in [the diff](https://github.com/pelias/schema/pull/473/commits/933318cc5bba96be52de7af0de2055fad05e44e8):
1. we lose the ability to index `alpha2` values in these fields and for it to 'just work'
2. we lose the ability to search by `alpha2` codes on the non-ngram field, for most, but not all, country codes.

FWIW I would happily revert that last commit once the analyser mess is sorted out but for now this is most likely better/safer.

orangejulius:
Woah...ok I was completely misunderstanding how this PR operated

I _thought_ that the `country_code` synonyms file was _only_ being applied to the `parent.country_a` field.

If it's not then...can we make it operate that way? I don't see any reason for the `country_code` synonyms, even a limited subset of them, to be applied to _any_ field other than `parent.country_a`. Not `parent.*` and certainly not `name.*`.

Remember, there's a very specific and limited type of query we're trying to fix here:  _autocomplete_ queries that end in a two letter country code (one that does _not_ prefix the 3 letter code).

 Any other type of query already works, and doesn't need to be considered. So, for example:
- a search query with any 2 or 3 letter country code works fine
- an autocomplete query with a 3 letter country code works fine.

Here's the simplest example of a currently failing case from the acceptance-tests:

https://github.com/pelias/acceptance-tests/blob/9a501fa7ac6befa68db3fa47914280bc22f9490f/test_cases/autocomplete_iso2_to_iso3.json#L7-L24

I also think we only need to apply the synonyms at index time. Replacing any instance of one of those 2 letter country codes with the 3 letter in a querystring would be unnecessary as far as I can tell.

I think this means we only need to consider the `parent.country_a.ngram` field as well.

Hopefully that clears things up and allows us to significantly reduce the impact that this PR has on our more common analyzers.

missinglink:
It's really not ideal, I highlighted these issues exactly one month ago in https://github.com/pelias/schema/pull/472 BTW

I would also prefer to fix the existing issues mentioned in that PR description but it will require significantly more engineering hours than we estimated for the task.

missinglink:
If we're *only* concerned about solving that one testcase we can probably get away with *only* fixing this part of my comment:

> For the `parent.*.ngram` fields we index with `"analyzer": "peliasIndexOneEdgeGram"`, which means that the synonyms will also be added to the `name.*` fields, this could be avoided if we also had individual ngram analyzers for each of these sub-fields which were different from the main `peliasIndexOneEdgeGram` analyzer.

... although for simplicity sake we could only do this for `country` and not the other admin fields for now.
it would require making a new `peliasIndexCountryAbbreviationEdgeGram` analyzer (that's a mouthful), but could be accomplished with limited additional engineering hours.

It would also allow us to specify a different search-time analyzer which didn't do synonym substitution at query-time.

That would mean that this part of the query would match:

Screenshot 2021-02-12 at 17 09 46

But it's worth noting that **queries for the alpha2 code which target the name.✱ fields will not work**, this may or may not be an issue, it depends on the query.

I know some of them, at least in the past, target that the `name.*` field as a `MUST` condition, although I don't know how to generate one of those off the top of my head.

Again if it's only that one testcase then this isn't an issue.

test/settings.js
```diff
@@ -50,23 +50,6 @@ module.exports.tests.analysis = function(test, common) {
 
 // -- analyzers --
 
-// this multiplexer filter provides all the synonyms used by the peliasAdmin analyzer
```
missinglink, line 53: these tests just got moved down the file.
they were listed under `analyzer` when they're really a `token_filter`

missinglink:
I've refactored this PR to add the new analyzers mentioned in my comments, this is the cleanest solution as there is now zero synonym crossover to other fields.

Tested on `Jamaica` locally and works great, I've kicked off a global build.

orangejulius:
Awesome. Definitely a bit more hassle and complexity added to our schema, but since it does look like only the `parent.country_a` needs these synonyms, it's probably the best choice :)

missinglink:
Another before test case I hope we can fix 
Screenshot_20210216-190931

orangejulius:
Ran this against our full planet acceptance test setup and there are no regressions!

PR #480, opened 2021-10-13
add dedicated concordance field

missinglink:
ping! @pelias/contributors this PR is a discussion with code attached 🚀 

this year has seen some work around recording and exposing 'concordances' (the WOF term for foreign key references).
these concordances are valuable to organisations who also use the foreign ID system and would like an easy way of joining Pelias GIDs with other datasets.

Screenshot 2021-10-13 at 13 53 33

the existing implementation works great, looking at [Germany in WOF](https://pelias.github.io/compare/#/v1/place?ids=whosonfirst%3Acountry%3A85633111) you can see it returns a treasure trove of useful concordances in the `addendum`.

one problem we've identified with using the addendum is that it's (by definition) only semi-structured and comes without many guarantees of correctness or availability.

what would be better is if concordances were more structured and formalised within Pelias so that they could be considered a public API which integrators could rely upon for a 'crosswalk' between datasets.

this PR would potentially open the door for that, it could be combined with a PR to `pelias/model` to perform the validation.
the validation rules would need a little thought, but things like casing, delimiters, abbreviations, collisions, etc would need to be considered.

there is also a secondary concern (beyond simply displaying the information), which is that users may also wish to search on these values, this is certainly never going to be possible with the addendum.

introducing a new parameter would need a bit more discussion but what comes to mind is the `/v1/place` endpoint could support `concordance` lookup, either via the existing `?ids=` param or a new one.

thoughts?

### DIFF
--- a/mappings/document.js
+++ b/mappings/document.js
@@ -171,6 +171,9 @@ var schema = {
     population: multiplier,
     popularity: multiplier,
 
+    // concordance (indexed foreign key references)
+    concordance: hash,
+
     // addendum (non-indexed supplimentary data)
     addendum: hash
   },
@@ -195,6 +198,12 @@ var schema = {
         search_analyzer: 'peliasQuery',
         similarity: 'peliasDefaultSimilarity'
       }
+    }
+  },{
+    concordance: {
+      path_match: 'concordance.*',
+      match_mapping_type: 'string',
+      mapping: keyword
     }
   },{
     addendum: {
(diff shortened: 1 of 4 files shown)


orangejulius:
Yeah, this makes a lot of sense, and I really like the idea of querying for concordances on the place endpoint. What do you think would be a good query format for that?

My memory is a bit hazy, but I think we should be able to query on those `keyword` fields easily, right? We don't need to do anything else: aggregations, keywords, or regular full text search.

PR #483, opened 2022-03-17
ES8 support - remove deprecated include_type_name parameter

jnorkus:
In ES8 the include_type_name parameter is no longer supported and causes an error when creating an index.

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html

### DIFF
--- a/integration/dynamic_templates.js
+++ b/integration/dynamic_templates.js
@@ -49,8 +49,7 @@ function nameAssertion( analyzer, common ){
     suite.assert( done => {
 
       suite.client.indices.getMapping({
-        index: suite.props.index,
-        include_type_name: false
+        index: suite.props.index
       }, (err, res) => {
 
         const properties = res[suite.props.index].mappings.properties;
@@ -88,8 +87,7 @@ function phraseAssertion( analyzer, common ){
     suite.assert( done => {
 
       suite.client.indices.getMapping({
-        index: suite.props.index,
-        include_type_name: false
+        index: suite.props.index
       }, ( err, res ) => {
 
         const properties = res[suite.props.index].mappings.properties;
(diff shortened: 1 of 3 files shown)


missinglink:
Thanks for the PR, we aren't officially supporting 8 just yet but we'll revisit this PR when we do.

The `_type` deprecation has been specially handled in our codebase for some time, it used to be required, then optional and now rejected, so I'll be happy to finally see the end of it 😄

missinglink:
related: https://github.com/pelias/schema/pull/401

jnorkus:
@missinglink Is there an issue/roadmap to support ES8? My colleague recently upgraded the cluster and it's not trivial to go back. Now I'm considering whether I should make necessary changes to support it. Anything else to be aware of?

missinglink:
Ugh no, TBH I didn't actually realize it was released a couple weeks ago.

There are loads of companies using Pelias and only three developers actively contributing to the maintenance, so things like this take some time.

It would be helpful if you could put together a list of the important differences and breaking changes for us to review.

missinglink:
Also worth mentioning that the license change makes some organizations a little uncomfortable.

I personally don't think it's a problem but it's a non-technical issue which will need to be smoothed over before we make 8 the default.

see: https://github.com/pelias/pelias/issues/926

missinglink:
**If** omission of a `_type` is the only real blocker **and** it allows the code to interoperate between versions 7 **and** 8 then I'd advocate making that configurable or inferable so we don't need to make it a breaking change.

jnorkus:
I managed to get index creation errors out of the way (there's more than just this parameter) but it seems that the importers are not compatible with this new index.

missinglink:
The importers all use `pelias/model` which still sets the `_type` (it's configurable), so we could change that so if you set it to something like `false` in the config then it is omitted.

see: https://github.com/pelias/model/pull/123

I suspect this `_type` deprecation is the main problem with migration and suspect that us using an older deprecated version of the official elasticsearch J's client isn't helping matters.

gnanakeethan:
@missinglink I am running this against Open Search 2.3 and removing the line in pelias/model#123 resolves the issue.

michaelkirk:
Fixed in https://github.com/pelias/schema/pull/490

PR #484, opened 2022-05-30
Update check_version.js

MALKARAJ:
change on line number 11 -> server name is incorrect

---

---

---


missinglink:
The word 'semver' is actually correct here, it's short for [semantic version](https://semver.org/#semantic-versioning-specification-semver)

PR #485, opened 2022-06-22
fix(typo): updating dockerfile to copy and install  packages in the w…

pushkar-geospoc:
…orkdir

---

---

---


pushkar-geospoc:
> Good catch, what was it doing before? I guess it was still working due to copying into the [CWD](https://en.m.wikipedia.org/wiki/Working_directory) which was equivalent?

yes due to CWD it was working but the dockerfile was inconsistent so think is better to make it consistent

PR #489, opened 2023-07-03
CI: Remove Node.js 12 from build matrix

orangejulius:
We normally try to keep all our CI configurations identical, but the schema repository is a bit of a special case since it has to do so many extra things around Elasticsearch.

Apparently, one of those extra things, `npm install -g npm`, now installs a new enough version of NPM that it doesn't work on Node.js 12.

We are removing Node.js 12 support globally soon anyway in https://github.com/pelias/pelias/issues/950,
so we might as well remove it here to make all the tests pass again.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x]
     steps:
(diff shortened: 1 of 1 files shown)


orangejulius:
Actually, I'd like to add `npm install -g npm` globally to all our CI eventually. It's a good best practice to insure NPM itself is working as well as it can.

We should probably make a ticket to track making this change, as well as a few others, across all our CI configs. As I recall they are all drifting a bit as we explore new things like how to enable unit tests on external pull requests, etc.

PR #492, opened 2023-07-16
Remove JDK configuration from CI

orangejulius:
For quite some time we have configured a specific version of the JDK in our CI.

As far as I understand it, modern versions of Elasticsearch use a bundled version of the JDK so this configuration may no longer be relevant.

On a practical/security note: it also looks like [the scripts](https://github.com/michaelklishin/jdk_switcher) used to switch JDK versions are over 4 years old, and just calling a bash script from a 3rd party repository is a bit scary.

I don't see any changes due to the removal of this option, is it possible we don't need it anymore? @missinglink what do you think?

### DIFF
--- a/.github/workflows/_integration_tests.yml
+++ b/.github/workflows/_integration_tests.yml
@@ -9,7 +9,6 @@ jobs:
           - ubuntu-22.04
         node-version: [18.x, 20.x, 22.x]
         es-version: [7.6.1]
-        jdk-version: [oraclejdk11]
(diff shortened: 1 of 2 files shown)


missinglink:
Yeah I'm fine to remove this.

The original reason was in order to be able to test different versions of the JDK in a split-test production environment.

I think that can also be achieved through custom Docker files.

orangejulius:
I found this old PR and updated it. Seems like it's still a good thing to merge.

PR #495, opened 2024-03-11
integration: merge elastictest into this module

missinglink:
I've archived https://github.com/missinglink/elastictest and pulled the `Suite.js` file into this repo.

It turned out no other repos used that npm module so including the src in this module directly makes maintenance and updates of that code easier since it's in the same place as the implementation.

cc/ @michaelkirk, I will pull your PR commits over here too.


michaelkirk:
Yay, good call!

PR #496, opened 2024-09-26
Add "«", "»" to blacklist characters list

SiarheiFedartsou:
:wave:

We faced an issue that our Pelias instance couldn't find `ТРЦ «Respublika Park»` (https://www.openstreetmap.org/way/293888317) via queries which don't contain  "«", "»"  symbols(e.g. I expect `Respublika Park` to be almost perfect match for this place). After some experimentations I narrowed it down to the fact that Pelias's ElasticSearch schema doesn't consider these symbols as blacklist, i.e. it becomes part of token (now we get `«Respublika` and `Park»`  as separate tokens). So I propose to add it to list of blacklisted characters to fix the issue. Such symbols are quite popular in some regions, see this wiki article https://en.wikipedia.org/wiki/Guillemet

---
"«", "»" symbols are quite popular delimiters in some languages. 

---
Added "«", "»" symbols to blacklist characters list

---

I updated tests + it can be tested manually via bootstrapping ES instance.


missinglink:
Seems reasonable, we actually have those symbols listed as quote characters in https://github.com/pelias/parser/blob/7d28b15f3cebee2d5739266f82cb3f0daa0f17f4/tokenization/split_funcs.js#L10

missinglink:
Is this good to merge or would you like to add additional symbols?

missinglink:
I took the liberty to clean up your commits, add some additional symbols and clean up some code, hope you don»t mind 🙏 

Thanks!

SiarheiFedartsou:
@missinglink I definitely don't, thanks!

missinglink:
I screwed up the commit message so the docker images only just got published now after I fixed it 🤦‍♂️.
https://github.com/pelias/schema/releases/tag/v7.1.0

PR #497, opened 2024-11-24
Bump NodeJs version on CI, fix compatibility with Node >20 in tests

SiarheiFedartsou:
:wave:

I was working on something and realised that `create_index.js` doesn't work on modern Node versions... Also decided to update CI to run on Node 18/20/22 and removed Node 16 (let me know if you think it is still needed, but IMO it is too outdated )

### DIFF
--- a/scripts/create_index.js
+++ b/scripts/create_index.js
@@ -38,5 +38,5 @@ client.indices.create(req, (err, res) => {
     process.exit(1);
   }
   console.log('[put mapping]', '\t', indexName, res, '\n');
-  process.exit(!!err);
+  process.exit(0);
 });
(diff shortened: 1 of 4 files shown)


scripts/create_index.js
```diff
@@ -38,5 +38,5 @@ client.indices.create(req, (err, res) => {
     process.exit(1);
   }
   console.log('[put mapping]', '\t', indexName, res, '\n');
-  process.exit(!!err);
```
SiarheiFedartsou, line 41: In the case of error we do `process.exit(1)` some lines above. This was incompatible with Node >20, because it checks that argument is number and it is not, then throws...

SiarheiFedartsou:
@missinglink please take a look when you have a chance 🙏🏻

missinglink:
Thanks 🙏

PR #498, opened 2024-11-26
Use ICU tokenizer to improve some Asian languages support

SiarheiFedartsou:
:wave: 

Disclaimer: I am not native to Thai, Chinese, Japanese or any languages I am adding support for here, but this change still seems to be reasonable. 

---
At the moment we use simple regex based tokenizer which simply splits strings using whitespaces and dashes as separator (`[\s,/\\-]+`). While it perfectly works for most of Western languages it fails in some Asian ones. Mainly it is because they don't actually use spaces to separate words (!). This change should address this via using [icu_tokenizer](https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu-tokenizer.html) which uses different complicated technics to solve it. 

Worth saying that this solution is still not perfect, e.g. for Thai there is separate `thai_tokenizer` (https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-thai-tokenizer.html), but icu one still works better than what we are using right now...

---
1. Changed tokenizer to icu_tokenizer
2. Added workaround for ampersand - icu_tokenizer filters it out...

---
I added a couple of integration tests with examples of queries which were tokenized as single token in the past.


SiarheiFedartsou:
@missinglink please take a look when you'll have a chance 🙏🏻 
Also do you think there are other places where we would need to make similar changes to have better support for such languages?

integration/analyzer_peliasStreet.js
```diff
@@ -1,4 +1,5 @@
 // validate analyzer is behaving as expected
+const { assert } = require('@hapi/joi');
```
missinglink, line 2: I think this unused import can be removed?

integration/analyzer_peliasQuery.js
```diff
@@ -49,6 +49,16 @@ module.exports.tests.functional = function(test, common){
     assertAnalysis( 'place', 'Toys "R" Us!', [ 'toys', 'r', 'us' ]);
     assertAnalysis( 'address', '101 mapzen place', [ '101', 'mapzen', 'place' ]);
 
+     // complicated tokenization for some Asian languages
+     assertAnalysis('thai_address1', 'ซอยเพชรบุรี๑', ['ซอย', 'เพชรบุรี1'] );
```
missinglink, line 53: Can you explain a little more about the segmentation here?

Are these token groups from research/knowledge or just a copy->paste of what was generated?

The number 1 on the end catches my eye and I'm curious how it appeared from `ซอยเพชรบุรี๑`

```
'ซอย', 'เพชรบุรี1'
```

settings.js
```diff
@@ -22,16 +22,16 @@ function generate(){
     "analysis": {
       "tokenizer": {
         "peliasTokenizer": {
-          "type": "pattern",
-          "pattern": "[\\s,/\\\\-]+"
```
missinglink, line 26: The previous pattern matched on `[whitespace, comma, forward/bash slashes and hyphen]`, is this still the case with the `icu_tokenizer`?

settings.js
```diff
@@ -175,6 +181,12 @@ function generate(){
         },
       },
       "filter" : {
```
missinglink, line 183: I wish this `AMPERSANDPLACEHOLDER` replacement wasn't necessary but understand why it is.

missinglink (commented):
This would be an interesting change, I'd like to continue investigation.

It will need significant testing since this is an absolutely core function of the system, even small regressions here will be hugely impactful.

It's promising that the `integration` tests all still pass, pointing to a possibility that this can be implemented in a backwards-compatible way.

I've added a few comments in the review.

Mostly I'd like to better understand the differences between the two tokenizers, any potential regressions/changes to the existing method, and enumerate the benefits of this work.

As it's a major change, it would be great to list at least a few queries which would improve with this change so other developers understand the value.

integration/analyzer_peliasQuery.js
SiarheiFedartsou: >Are these token groups from research/knowledge or just a copy->paste of what was generated?

Well, I used ChatGPT to generate them and then tools like Google Translate to validate, that these tokens indeed have separate meanings... Tbh it would be great to have someone native to one of those languages to validate 🤔  I will try to find one 😄 

>The number 1 on the end catches my eye and I'm curious how it appeared from ซอยเพชรบุรี๑

Thai language has its own digits (but they use Arabic ones as well btw) and `๑` is `1`. Btw we convert them already even without this tokenizer. Not sure at which step though. But good question is if this number should be a separate token, but IIRC it is how it works right now(if you input `foo1` it will generate single `foo1` token), so I decided to limit  scope of changes and may be will address it with separate PR. WDYT?

settings.js
SiarheiFedartsou: Good point, I'll try to check if we have relevant tests and if we don't, will write them :+1:

SiarheiFedartsou:
@missinglink thanks for the review! I'll come with more answers to your questions later.

integration/analyzer_peliasQuery.js
missinglink: and yeah, we should get a human to check a few examples, maybe beat to generate a few more before asking someone in different classes such as address, venue(s), streets, intersections, localities etc.

missinglink: this feature will go through a bunch of testing which might take a long time, flagging potential issues now I'll be much easier than finding them later, I'm fairly confident we can resolve any issues.

integration/analyzer_peliasQuery.js
```diff
@@ -49,6 +49,16 @@ module.exports.tests.functional = function(test, common){
...
+     assertAnalysis('thai_address2', 'ซอยเพชรบุรี๑foo', ['ซอย', 'เพชรบุรี1', 'foo'] );
+     assertAnalysis('thai_address3', 'บ้านเลขที่๑๒๓ถนนสุขุมวิทแขวงคลองตันเหนือเขตวัฒนา กรุงเทพมหานคร๑๐๑๑๐', ["บาน", "เลข", "ที123ถนน", "สุขุมวิท", "แขวง", "คลองตัน", "เหนือ", "เขต", "วัฒนา", "กรุงเทพมหานคร10110"]);
+     assertAnalysis('chinese_address', '北京市朝阳区东三环中路1号国际大厦A座1001室', 
+       ['北京市', '朝阳', '区', '东', '三', '环', '中路', '1', '号', '国际', '大厦', 'a', '座', '1001', '室']);
```
kenil-cheng, line 57: The correct splitting is:
北京市 - Beijing city
朝阳区 - The district
东三环中路 - East 3rd Ring Middle Road
1号 - number
国际大厦 - Building name
a座 - Block number
1001室 - Room number

integration/analyzer_peliasQuery.js
```diff
@@ -49,6 +49,16 @@ module.exports.tests.functional = function(test, common){
...
+     assertAnalysis('japanese_address', '東京都渋谷区渋谷２丁目２１−１渋谷スクランブルスクエア４階', ["東京", "都", "渋谷", "区", "渋谷", "2", "丁目", "21", "1", "渋谷", "スクランフル", "スクエア", "4", "階"]);
```
kenil-cheng, line 58: The correct splitting is:
東京都 - Tokyo city
渋谷区 - District
渋谷2 丁目 - Street
211 - Street number
渋谷スクランフルスクエア - Building: Shibuya Scramble Square
4 階 - Floor

integration/analyzer_peliasQuery.js
kenil-cheng: The full Chinese addresses are usually ...省 ...市 ...区 ...路 ...号 building_name ...座 ...楼 ...室

missinglink:
Thank you @kenil-cheng for your time.

SiarheiFedartsou:
> Possibly stupid question, but does Elastic support the regex [word boundary character \b](https://www.regular-expressions.info/wordboundaries.html) and is that unicode-aware?

I've tried to simply extend original regex with it and got error (smth about it doesn't know what `\b` is :( )

integration/analyzer_peliasStreet.js
```diff
@@ -22,6 +22,16 @@ module.exports.tests.analyze = function(test, common){
     assertAnalysis( 'remove_ordinals', '1st 2nd 3rd 4th 5th', ['1','2','3','4','5'] );
     assertAnalysis( 'remove_ordinals', 'Ast th 101st', ['ast','th','101'] );
 
+    // complicated tokenization for some Asian languages
+    assertAnalysis('thai_address1', 'ซอยเพชรบุรี๑', ['ซอย', 'เพชรบุรี1'] );
+    assertAnalysis('thai_address2', 'ซอยเพชรบุรี๑foo', ['ซอย', 'เพชรบุรี1', 'foo'] );
+    assertAnalysis('thai_address3', 'บ้านเลขที่๑๒๓ถนนสุขุมวิทแขวงคลองตันเหนือเขตวัฒนา กรุงเทพมหานคร๑๐๑๑๐', ["บาน", "เลข", "ที123ถนน", "สุขุมวิท", "แขวง", "คลองตัน", "เหนือ", "เขต", "วัฒนา", "กรุงเทพมหานคร10110"]);
```
dzviagin, line 28: My Thai mate suggested 
> ["บ้าน", "เลขที่", "123", "ถนน", "สุขุมวิท", "แขวง", "คลองตันเหนือ", "เขต", "วัฒนา", "กรุงเทพมหานคร", "10110",]
apart from different split, it's visible that first letter was changed from บ้ to บ in your case.

SiarheiFedartsou: Thanks @dzviagin ! I will take a look, likely it is a result of removal of diacritic symbols, which existed even before this PR...

SiarheiFedartsou:
Hey @missinglink WDYT we would need to do in order to eventually get it merged?

missinglink:
Hi @SiarheiFedartsou 

This is a significant change to probably the most core function in the system, which is always going to be a difficult thing to merge confidentiality.

I see a few paths to adoption:
1. Merge it as-is and hope nothing breaks.
2. Invest some significant time testing it against global datasets and queries, then merge it.
3. Merge it behind a beta flag, encourage adoption and iterate based on feedback to a stable release.
4. Keep it in a fork and encourage adoption, as above.

We need to be very careful to consider all the current users across business/local govt etc who will be impacted by the change (positively or negatively).

For this reason I feel like 1. is potentially a bit reckless, maybe you can assure me it's not going to affect existing Latin queries?

Option 2. is a good idea but requires potentially a lot of time invested by the core team to spin up global builds and test them, a build generally takes most of a day, testing also takes time, iterating on this process takes a lot of time.

Option 3. is the easiest solution to getting the code into the master branch, this would allow users to test it out and provide feedback. The issue being that it's going to be enabled by a small percentage of users, although it would also make 2. easier for us to automate. The code would potentially be messier during the beta period (code duplication).

One clean way of achieving this is to have a new `pelias/config` variable which is read when the index is created, call it `icuTokenizer: boolean` or similar which tells this lib to use the icu tokenizer code path.

Option 4. is much like 3 except it's even less likely that users will find the fork and use it as a replacement for the master branch, reducing the amount of feedback we receive.

It seems to me that 3. would be something we could accept immediately and would make 2. easier to automate.

Please let me know your thoughts and if I've missed something, or if you think I'm overstating any potential regressions.

cc/ @orangejulius @Joxit for their opinions

SiarheiFedartsou:
@missinglink yes, I agree. While I am quite sure it won't affect existing Latin queries from business logic perspective, I am not sure in other perspectives: e.g. I would expect this tokenizer to be slower and it is not clear for how much.

Okay, I'll probably try to invest some time to investigate how this change can be made optional. Thanks!

missinglink:
Regarding the performance impact:

I suspect that the indexing performance to be negligible yet considering the scale of 500+ million documents it could become an issue, let's keep an eye on this but it's likely not a big deal to add 1-2% to build times considering the benefits.

For queries (where we also run the tokenizer) I also suspect it will be negligible, but yeah let's keep an eye on that too, particularly for autocomplete which is latency sensitive.

We (Geocode Earth) have the ability to split test our deployments and measure latency differences between API traffic sent to each. It's something we'd do once we were sure the quality wasn't affected but would allow us to scientifically measure the latency impact on the scale of tens of millions of queries.

SiarheiFedartsou:
@missinglink may I ask you to take a look again please? :) 

What I have now:
- I added `icuTokenizer` flag to config (I assume it is okay to simply add it to `configValidation` and don't update https://github.com/pelias/config ?)
- I rely on this flag in `settings.js` to generate proper schema. I tried to avoid copy-paste as much as possible.
- I extended CI, so that we run tests for both icuTokenizer = true/false
- I updated tests, so that for some tests we have separated branches for icuTokenizer = true/false

missinglink:
Looks good to me:

```bash
/code/p/schema pull/498/merge = *1 ❯ 2>&1 npm run integration | grep -i thai
    ✔ thai_address
    ✔ thai_address1
    ✔ thai_address1
```

```bash
/code/p/schema pull/498/merge = *1 ❯ jq -n '{ schema: { icuTokenizer: true } }
/code/p/schema pull/498/merge = *1 ?1 ❯ export PELIAS_CONFIG=$(pwd)/config-icu.json
/code/p/schema pull/498/merge = *1 ?1 ❯ 2>&1 npm run integration | grep -i thai
    ✔ thai_address
    ✔ thai_address1
    ✔ thai_address2
    ✔ thai_address3
    ✔ thai_address1
    ✔ thai_address2
    ✔ thai_address3
```

missinglink:
Does this PR include https://github.com/pelias/schema/pull/501 or do we need to merge that seperately?

SiarheiFedartsou:
> Does this PR include #501 or do we need to merge that seperately?

It doesn't. Let's may be merge it separately?

missinglink:
I really wish we didn't need `ampersand_mapper` and `ampersand_replacer`, removing them would make this much cleaner, unfortunately I don't see any way of achieving this 🤷 

I'm happy to merge this tomorrow unless we get any additional feedback from the linked contributors before then.

missinglink:
Thanks for contributing this.
I added https://github.com/pelias/schema/pull/498/commits/eefcaaad123233652b45d6134a8e61345e0b6bb6 to do a little cleanup before merging 🙇

orangejulius:
Hey sorry, I'm late to this PR but I'd like to request some changes.

We've extended considerable effort to ensure that the Github actions templates across all our Pelias repos are as identical as possible. Would it be possible to adjust the testing done in this PR so it does not work against that?

I'm also a little concerned about having completely dupilicate fixtures and expected test values. Those file are already a bit of a pain to keep up to date, and now we'll have two of them.

Do we think that having multiple tokenizers is something we would want to have in the Pelias codebase long term? If not, let's just keep this on a branch and test it until it's better?

PR #499, opened 2024-11-27
Use ES 8.x on CI

SiarheiFedartsou:

---

---

---


SiarheiFedartsou:
I will debug it in my fork first :)

PR #507, opened 2025-03-10
feat: Allow variable index names for pelias docker

arnesetzer:

---

Closes https://github.com/pelias/docker/issues/371 which allows the docker command  `pelias elastic stats` to use a different index name than the hardcoded ("pelias"). 

---
Moves the logic of the command `pelias elastic stats` from the docker repo to this repo to be able to use the command with other index names.

---

1. Change `pelias.json` like
```json
"schema": {
    "indexName": "everything except pelias"
},
"api": {
    "indexName": "everything except pelias"
}
```
2. Import some data
3. Execute `node script/check_stats`
or
1. `docker build -t differentName/pelias-schema -f Dockerfile . `
2. Modify the docker-compose.yml to
```yaml
 schema:
    image: differentName/pelias-schema
```
3. Change `pelias.json` like
```json
"schema": {
    "indexName": "everything except pelias"
},
"api": {
    "indexName": "everything except pelias"
}
```
4. Import some data
5. Execute `pelias elastic stats`

### DIFF
--- a/scripts/check_stats.js
+++ b/scripts/check_stats.js
@@ -1,11 +1,13 @@
 const config = require('pelias-config').generate();
 const es = require('elasticsearch');
 const client = new es.Client(config.esclient);
+const targetIndex = getTargetIndex();
 
-async function stats() {
+
+async function stats(targetIndex) {
     try {
         const response = await client.search({
-            index: config.schema.indexName,
+            index: targetIndex,
             body: {
                 aggs: {
                     sources: {
@@ -29,7 +31,7 @@ async function stats() {
             request_cache: true,
             maxRetries: 1,
         });
-        console.log("Results for index \""+config.schema.indexName+"\":")
+        console.log("Results for index \"" + targetIndex + "\":")
         console.log(JSON.stringify(response, null, 2));
         process.exit(0);
     }
@@ -37,5 +39,12 @@ async function stats() {
         console.error(err);
         process.exit(!!err);
     }
+};
+
+function getTargetIndex() {
+    if (process.argv.length > 2 && ['--api', '-a'].indexOf(process.argv[2]) > -1) {
+        return config.api.indexName; //If none is set the value from pelias/config/config/defaults.json will be used ('pelias')
+    }
+    return config.schema.indexName;
 }
-stats();
+stats(targetIndex);
--- a/scripts/check_stats.js
+++ b/scripts/check_stats.js
@@ -1,0 +1,41 @@
+const config = require('pelias-config').generate();
+const es = require('elasticsearch');
+const client = new es.Client(config.esclient);
+
+async function stats() {
+    try {
+        const response = await client.search({
+            index: config.api.indexName,
+            body: {
+                aggs: {
+                    sources: {
+                        terms: {
+                            field: 'source',
+                            size: 100
+                        },
+                        aggs: {
+                            layers: {
+                                terms: {
+                                    field: "layer",
+                                    size: 100
+                                }
+                            }
+                        }
+                    }
+                },
+                size: 0,
+            },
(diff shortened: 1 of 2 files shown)


scripts/check_stats.js
```diff
@@ -0,0 +1,41 @@
+const config = require('pelias-config').generate();
+const es = require('elasticsearch');
+const client = new es.Client(config.esclient);
+
+async function stats() {
+    try {
+        const response = await client.search({
+            index: config.api.indexName,
```
missinglink, line 8: I think this should be `config.schema.indexName`?

- https://github.com/pelias/schema/blob/master/scripts/create_index.js#L29
- https://github.com/pelias/config/blob/master/config/defaults.json#L73

missinglink (changes requested):
With that one minor change it works great:

```bash
docker compose run --rm -it schema node scripts/check_stats.js
```

```bash
{
  "took": 157,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 10000,
      "relation": "gte"
    },
    "max_score": null,
    "hits": []
  },
  "aggregations": {
    "sources": {
      "doc_count_error_upper_bound": 0,
      "sum_other_doc_count": 0,
      "buckets": [
        {
          "key": "openaddresses",
          "doc_count": 1389013,
          "layers": {
            "doc_count_error_upper_bound": 0,
...
`

**Comment by missinglink** [comment]
Can you please also change the other line in that file:
```
console.log(`Results for index ${config.schema.indexName}`)
```

scripts/check_stats.js
arnesetzer: I wasn't sure about this. In my opinion you want to get the stats from your active queried index. But I guess you can argue for both sides.

missinglink: This documentation is actually wrong, I'll fix it:

https://github.com/pelias/api/blob/6037d952d2ed652f883803cf9a590fe3ad968173/README.md?plain=1#L71

The API reads from `config.schema.indexName` too, so using the same variable will guarantee that they will point at the same index.

missinglink: I see the confusion now, we have two variables for essentially the same thing.
@arnesetzer is there a valid scenario where these two variables might differ?

arnesetzer: "Updating" the elasticsearch database. We use the old index (`api.indexName`) for requests while importing data to the new index (`schema.indexName`). After the importing processes are finished we just need to swap the index and you have almost no downtime.

missinglink: I recommend [aliases](https://www.elastic.co/guide/en/elasticsearch/reference/current/aliases.html) for this scenario, we used it in the past with great success.

basically it acts like a symlink to an index so you can switch `pelias->pelias-2025-01` to  `pelias->pelias-2025-02` with zero downtime by sending a command to Elasticsearch.

missinglink: sorry, I need to chat with Julian about this some more before we merge it, I wasn't totally up-to-date on the fact we had two variables and the purpose of that.

missinglink:
So.. regarding `api.indexName` vs. `schema.indexName` it sounds like we have them *both* for exactly the scenario you described, where multiple indices are generated in the same ES server and there is flexibility to configure them independently. https://github.com/pelias/api/pull/1691#issuecomment-2714739952

As to *which* of the two we use here.. if we set it to `api.indexName` then you can't use it to check the stats of new builds before they are promoted to production.

If we set it to `schema.indexName` then you can't use it to see the active production stats.

Would it make sense to let the script accept an `argv` argument which instead so it could be used for *either*?

